PR #63, opened 2020-05-11
Remove duplicated "Phase 2 Phase 2" text

ethers:
There was a double "Phase 2 Phase 2" in the text.  Took the opportunity to improve the descriptions.

### DIFF
--- a/src/pages/Landing/Phases/index.tsx
+++ b/src/pages/Landing/Phases/index.tsx
@@ -92,7 +92,7 @@ export const Phases = (): JSX.Element => {
               </Text>
               <Text className="mt20">
                 <PhaseTitle>Phase 1 </PhaseTitle>
-                adds more data capacity to eth2.
+                adds data capacity to eth2.
               </Text>
               <Text className="mt20">
                 <PhaseTitle>Phase 2 </PhaseTitle>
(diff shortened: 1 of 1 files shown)


src/pages/Landing/Phases/index.tsx
```diff
@@ -92,12 +92,11 @@ export const Phases = (): JSX.Element => {
               </Text>
               <Text className="mt20">
                 <PhaseTitle>Phase 1 </PhaseTitle>
-                handles adding and storing the data associated with eth2.
+                adds more data capacity to eth2.
```
CarlBeek, line 95: I am not sure I agree with the wording here. In particular "more" was there already data capacity in phase 1?

ethers: Right, "more" has been deleted.  "storing the data associated with eth2" was removed because that sounds like data on who are the validators in eth2, their stakes, etc, and as you know, most of it is Phase 0.

samajammin:
Thanks for this contribution @ethers!

I believe this PR is now outdated. As part of #300 & #308 we re-vamped the site content to drop the phase terminology.

Feel free to re-open if you think there's something still missing.

PR #72, opened 2020-05-18
Update price per validator to 32, cleanup related usages

### LINKED ISSUE #34: Separate PRICE_PER_VALIDATOR for mainnet & testnet
There will probably be separate `PRICE_PER_VALIDATOR`s for mainnet & testnet. Ideally this would be supported.

It is probably better to represent this as a quotient or factor as the deposit contract will compensate for this.

For example, if validators are depositing 3.2 ETH/validator, the deposit contract will store their deposit as 32 ETH, but the contract's balance will only increase by 3.2 ETH. Using a quotient, it is easier to determine represent this amount in the interface for total deposits etc.

Battenfield:
- Updates `PRICE_PER_VALIDATOR` from 3.2 to 32
- Cleans up usages using `PRICE_PER_VALIDATOR`

We should deploy a new test Deposit Contract that reflects 32ETH per validator and update the `CONTRACT_ADDRESS` in this PR before merging

Fixes #34

### DIFF
--- a/src/pages/Acknowledgements/pageContent.tsx
+++ b/src/pages/Acknowledgements/pageContent.tsx
@@ -40,9 +40,9 @@ export const pageContent = {
     title: 'Signing up',
     content: (
       <Text size="large" className="my10">
-        To become a validator on eth2, you need to deposit 
-        {PRICE_PER_VALIDATOR}
-        ETH per validator that you wish to run. This process cannot be reversed.
+        To become a validator on Ethereum 2, you need to deposit{' '}
+        {PRICE_PER_VALIDATOR} ETH per validator that you wish to run. This
+        process cannot be reversed.
       </Text>
     ),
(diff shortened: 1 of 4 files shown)


src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -40,9 +40,9 @@ export const pageContent = {
     title: 'Signing up',
     content: (
       <Text size="large" className="my10">
-        To become a validator on eth2, you need to deposit 
-        {PRICE_PER_VALIDATOR}
-        ETH per validator that you wish to run. This process cannot be reversed.
+        To become a validator on Ethereum 2, you need to deposit{' '}
```
CarlBeek, line 43:
```suggestion
        To become an eth2 validator, you need to deposit{' '}
```

CarlBeek:
Resolved in #90

PR #73, opened 2020-05-19
Generate keys windows commands

### LINKED ISSUE #67: Windows commands for Key Generation page
We need the windows version of our Mac/Linux instructions for generating validator keys.
- Dependency instructions and links
- terminal/PowerShell commands
- specify python version

Screen Shot 2020-05-13 at 10 28 22 PM

Battenfield:
- Adds in windows commands from @qnou 

Fixes #67


CarlBeek:
Indeed it does look good, and I'm happy to merge for now, but I'd like to note that for someone who is not a dev, the windows commands are not very friendly. They need to install, chocolatey, git, pip and python all via Powershell, which I think not many people (including devs) are familiar with.

PR #79, opened 2020-05-25
Update BLS to v2 with H2C v7

CarlBeek:
As per the final version of BLS Eth2 is going to use ([Eth2 spec v0.12](https://github.com/ethereum/eth2.0-specs/releases/tag/v0.12.0)). This PR updates BLS to the new Herumi version (via chainsafe package) thereby supporting [BLS v2](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02) and [Hash-to-curve v7](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@chainsafe/bls": "^1.0.0",
+    "@chainsafe/bls": "^2.0.0",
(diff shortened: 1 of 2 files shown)


CarlBeek:
Resolved by #90 in https://github.com/ethereum/eth2.0-deposit/pull/90/commits/291bce81df3886d4e2fcd894d6e1b21eab4d6041

PR #83, opened 2020-06-02
Verifies SSZ encoding with @chainsafe/ssz

### LINKED ISSUE #61: use @chainsafe/ssz
You should verify th ssz parameters for the user using th `@chainsafe/ssz`

CarlBeek:
At the moment, the signature is verified against the root provided by the deposit-cli. This is unsafe if the root is not correctly calculated, the message that is signed could be incorrect.

In addition to the above, the deposit data json checker was brittle.

This addresses the above by using @chainsafe/ssz to verify the root as well as implementing the signature verification methods from the eth2-spec.

* Closes #61 
* Resolves ethereum/eth2.0-deposit-cli/issues/23 (Root re-verification)
* Required for ethereum/eth2.0-deposit-cli/pull/32 (`fork_version `)

@hwwhww, I'd like your review for the root checking & signature verification
@Battenfield, Please can you check my typescript.

### DIFF
--- a/src/pages/UploadValidator/validateDepositKey.ts
+++ b/src/pages/UploadValidator/validateDepositKey.ts
@@ -4,7 +4,48 @@
 import _every from 'lodash/every';
 import { initBLS } from '@chainsafe/bls';
 import { verifySignature } from '../../utils/verifySignature';
+import { verifyDepositRoots } from '../../utils/SSZ';
 import { DepositKeyInterface } from '../../store/reducers';
+
+const validateFieldFormatting = (
+  depositDatum: DepositKeyInterface
+): boolean => {
+  // check existence of required keys
+  if (
+    !depositDatum.pubkey ||
+    !depositDatum.withdrawal_credentials ||
+    !depositDatum.amount ||
+    !depositDatum.signature ||
+    !depositDatum.deposit_message_root ||
+    !depositDatum.deposit_data_root
+  ) {
+    return false;
+  }
+
+  // check type of values
+  if (
+    typeof depositDatum.pubkey !== 'string' ||
+    typeof depositDatum.withdrawal_credentials !== 'string' ||
+    typeof depositDatum.amount !== 'number' ||
+    typeof depositDatum.signature !== 'string' ||
+    typeof depositDatum.deposit_message_root !== 'string' ||
+    typeof depositDatum.deposit_data_root !== 'string'
+  ) {
+    return false;
+  }
+
+  // check length of strings
+  if (
+    depositDatum.pubkey.length !== 96 ||
+    depositDatum.withdrawal_credentials.length !== 64 ||
+    depositDatum.signature.length !== 192 ||
+    depositDatum.deposit_message_root.length !== 64 ||
+    depositDatum.deposit_data_root.length !== 64
+  ) {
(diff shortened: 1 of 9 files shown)


src/pages/UploadValidator/validateDepositKey.ts
```diff
@@ -4,8 +4,49 @@
...
+  depositDatum: DepositKeyInterface
+): boolean => {
+  // check existence of required keys
+  if (
+    !depositDatum.pubkey ||
+    !depositDatum.withdrawal_credentials ||
+    !depositDatum.amount ||
+    !depositDatum.signature ||
+    !depositDatum.deposit_message_root ||
+    !depositDatum.deposit_data_root
```
hwwhww, line 20: Field names inconsistent

1. launchpad's `deposit_message_root` is deposit-cli's `deposit_data_root`
2. launchpad's `deposit_data_root` is deposit-cli's `signed_deposit_data_root`

I'd suggest we modify deposit-cli side.

CarlBeek:
Thank for the review!

>Field name inconsistent issue.

I agree this should be changed on the deposit-cli side. As we discussed the other day, keeping naming consistent with the spec is a good idea.

> GENESIS_FORK_VERSION

This is addressed in #84 (which I made just after this PR). Ultimately, I'd like to redesign how network configuration is done, but I don't think my JS kung fu is up to snuff for this task. In much the same way we do in the CLI now, I'd like to have a config file with network, deposit amount, is_testnet, deposit contract address, fork version, etc built in.

src/pages/UploadValidator/validateDepositKey.ts
CarlBeek: Resolving, see: https://github.com/ethereum/eth2.0-deposit-cli/pull/32/commits/dbac6d236b1627221d8339f2ff8ec1d9ba1cb500

src/utils/envVars.ts
```diff
@@ -15,3 +15,12 @@ export const CONTRACT_ADDRESS = '0x4DC8B546b93131309c82505a6fdfB978D311bf45';
...
 export const IS_MAINNET = false;
+
+// eth2.0 specs constants
+export const BLS_WITHDRAWAL_PREFIX = Buffer.from('00', 'hex');
+export const DOMAIN_DEPOSIT = Buffer.from('03000000', 'hex');
+export const EMPTY_ROOT = Buffer.from(
+  '0000000000000000000000000000000000000000000000000000000000000000',
+  'hex'
+);
+export const GENESIS_FORK_VERSION = Buffer.from('00000000', 'hex');
```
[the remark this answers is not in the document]
Battenfield: We could pull this from the `.env` file for specialized deployments as well as, default it to `Buffer.from('00000000', 'hex');`

Battenfield:
> > GENESIS_FORK_VERSION
> 
> This is addressed in #84 (which I made just after this PR). Ultimately, I'd like to redesign how network configuration is done, but I don't think my JS kung fu is up to snuff for this task. In much the same way we do in the CLI now, I'd like to have a config file with network, deposit amount, is_testnet, deposit contract address, fork version, etc built in.

We could put all of these parameters in the `.env` and default them to their current `envVars.ts` values.

PR #88, opened 2020-06-09
Fixes #87: BigInt dependency that caused browser compatibility issues

### LINKED ISSUE #87: App crashes on Mobile and safari - Bigint error
#83 created a new app crashing bug on certain browsers
- Chrome Mobile
- Safari (desktop and mobile)

Screen Shot 2020-06-04 at 5 25 27 PM

CarlBeek:
#83 Added SSZ verification of the roots in the deposit-json file. Unfortunately, as was presented in #87, this broke compatibility with some browsers (those that do not support ES2020 yet).

@wemeetagain pushed some changes to @chainsafe/SSZ and its dependancies so that `BigInt`s are not called while calculating hash-tree-roots for sufficiently small ints.

This PR bumps the SSZ version, changes the local types, and performs range checks on the deposit amount.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,6 @@ module.exports = {
   globals: {
     Atomics: 'readonly',
     SharedArrayBuffer: 'readonly',
-    BigInt: true,
   },
   parser: '@typescript-eslint/parser',
   parserOptions: {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@chainsafe/bls": "^1.0.0",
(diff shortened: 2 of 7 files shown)


CarlBeek:
Good catch, thank you @Battenfield! I have updated the deposit contract to 32 ETH and compiled it from the solidity version.

[This is the version of the contract with the drain backdoor](https://github.com/CarlBeek/eth2.0-deposit-contract/tree/6f4b00b31d8b71032f942f7b375731a74a70748b) and it was [deployed here](https://goerli.etherscan.io/address/0x344b3a521ded954b4fa9ec8cc1d999631c998daa)

I have tested the deposit with the launchpad on this branch & the drain function works too.

PR #90, opened 2020-06-30
refactor envVars to allow for eth2 network configs

georgeweiler:
- simplifies the `.env` variables needed to run the app locally
- establishes a process for configuring the variables for various deployments


src/utils/envVars.ts
```diff
@@ -1,29 +1,23 @@
...
-export const GWEI = 1e9;
-export const MIN_DEPOSIT_AMOUNT = 1 * GWEI;
-export const MAX_DEPOSIT_AMOUNT = 32 * GWEI;
-export const PRICE_PER_VALIDATOR = MAX_DEPOSIT_AMOUNT / GWEI;
+// public
+export const ALETHIO_URL                = IS_MAINNET ? 'https://explorer.goerli.aleth.io/tx' : 'https://explorer.goerli.aleth.io/tx';
+export const ETHERSCAN_URL              = IS_MAINNET ? 'https://goerli.etherscan.io/tx' : 'https://goerli.etherscan.io/tx';
+export const FORTMATIC_KEY              = process.env.REACT_APP_FORTMATIC_KEY       || 'pk_test_D113D979E0D3508F';
+export const CONTRACT_ADDRESS           = process.env.REACT_APP_CONTRACT_ADDRESS    || '0x344b3a521ded954b4fa9ec8cc1d999631c998daa';
+export const ETH2_NETWORK_NAME          = process.env.REACT_APP_ETH2_NETWORK_NAME   || '';
```
hwwhww, line 14: Default to mainnet?
```suggestion
export const ETH2_NETWORK_NAME          = process.env.REACT_APP_ETH2_NETWORK_NAME   || 'mainnet';
```

src/utils/envVars.ts
```diff
@@ -1,29 +1,23 @@
...
+export const ETH_REQUIREMENT            = process.env.REACT_APP_ETH_REQUIREMENT     || 524288;
+export const PRICE_PER_VALIDATOR        = process.env.REACT_APP_PRICE_PER_VALIDATOR || 32;
+
+// BLS signature verification variables
+export const GWEI                       = 1e9;
+export const MIN_DEPOSIT_AMOUNT         = 1 * GWEI;
```
hwwhww, line 20: Nitpicking suggestion for readability:

```suggestion
export const ETHER_TO_GWEI              = 1e9;
export const MIN_DEPOSIT_AMOUNT         = 1 * ETHER_TO_GWEI;
```

.env.template
```diff
@@ -0,0 +1,18 @@
...
+REACT_APP_INFURA_PROJECT_ID=464bfe80298f4d0782f61h4b91de2dcb
+
+
+
+# true, false
+REACT_APP_IS_MAINNET=false
+# string
+REACT_APP_FORTMATIC_KEY=pk_test_D113D979E0D3508F
+# string
+REACT_APP_CONTRACT_ADDRESS=0x344b3a521ded954b4fa9ec8cc1d999631c998daa
```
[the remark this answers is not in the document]
hwwhww: If we add `REACT_APP_GENESIS_FORK_VERSION`, the Altona's GENESIS_FORK_VERSION is `Buffer.from('00000121', 'hex');`

.env.template
```diff
@@ -0,0 +1,18 @@
+# private strings
+REACT_APP_PORTIS_DAPP_ID=006cdb48-470e-466c-aa20-c4de6ebd10e7
```
CarlBeek, line 2: Are these actual secrets, or are they placeholder vars?

src/utils/envVars.ts
```diff
@@ -1,29 +1,23 @@
...
-  'hex'
-);
-export const GENESIS_FORK_VERSION = Buffer.from('00000000', 'hex');
-export const GWEI = 1e9;
-export const MIN_DEPOSIT_AMOUNT = 1 * GWEI;
-export const MAX_DEPOSIT_AMOUNT = 32 * GWEI;
-export const PRICE_PER_VALIDATOR = MAX_DEPOSIT_AMOUNT / GWEI;
+// public
+export const ALETHIO_URL                = IS_MAINNET ? 'https://explorer.goerli.aleth.io/tx' : 'https://explorer.goerli.aleth.io/tx';
+export const ETHERSCAN_URL              = IS_MAINNET ? 'https://goerli.etherscan.io/tx' : 'https://goerli.etherscan.io/tx';
```
CarlBeek, line 11: Point to mainnet config?

```suggestion
export const ALETHIO_URL                = IS_MAINNET ? 'https://explorer.aleth.io/tx' : 'https://explorer.goerli.aleth.io/tx';
export const ETHERSCAN_URL              = IS_MAINNET ? 'https://etherscan.io/tx' : 'https://goerli.etherscan.io/tx';
```

src/utils/envVars.ts
```diff
@@ -1,29 +1,23 @@
...
+export const ETH2_NETWORK_NAME          = process.env.REACT_APP_ETH2_NETWORK_NAME   || '';
+export const ETH_REQUIREMENT            = process.env.REACT_APP_ETH_REQUIREMENT     || 524288;
+export const PRICE_PER_VALIDATOR        = process.env.REACT_APP_PRICE_PER_VALIDATOR || 32;
+
+// BLS signature verification variables
+export const GWEI                       = 1e9;
+export const MIN_DEPOSIT_AMOUNT         = 1 * GWEI;
+export const DOMAIN_DEPOSIT             = Buffer.from('03000000', 'hex');
+export const EMPTY_ROOT                 = Buffer.from('0000000000000000000000000000000000000000000000000000000000000000', 'hex');
+export const GENESIS_FORK_VERSION       = Buffer.from('00000000', 'hex');
```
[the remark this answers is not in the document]
georgeweiler: yes, that makes sense! Can we default to `0x00000000` if `GENESIS_FORK_VERSION` is not provided? And if should `IS_MAINNET === true` override any custom `GENESIS_FORK_VERSION`?

.env.template
georgeweiler: These values are placeholders made to look like real values.

georgeweiler:
Hey @CarlBeek, when I was QA'ing these changes I noticed that the BLS Verification signature verification includes a check for the upper and lower deposit amounts compared against `depositDatum.amount`, which is always `32*GWEI` (as far as I know). 

I've updated the `MAX_DEPOSIT_AMOUNT` to use `PRICE_PER_VALIDATOR * GWEI * GWEI`, but now setting `PRICE_PER_VALIDATOR * GWEI` to anything less than 32 will cause the BLS check to fail, so the user is unable to upload their key file.


```if (
    depositDatum.amount < MIN_DEPOSIT_AMOUNT ||
    depositDatum.amount > PRICE_PER_VALIDATOR * GWEI
  ) {
    return false;
  }```

**Comment by CarlBeek** [comment]
This is an intended safety feature. The signature is over the amount too and its validity bounds are exactly as you described. This is another way of checking that the correct chain has been selected.

Basically, the the launchpad & CLI need to agree on the deposit sizes & on the fork version. On the CLI side, we implemented a new flag `--chain` which selects the chain (in the background it sets price/validator & the `GENESIS_FORK_VERSION`). Currently, `--chain witti` and `--chain mainnet` are supported. (After #91, we can add `altona` support.) 

In summary, we need to add an extra parameter to these network configs for the various cli network names and add them to the network configs. Note that mainnet is the default so we don't have to supply a cli chain-name for mainnet.

File: src/utils/envVars.ts (hunk 1 above)
Comment (georgeweiler) [reply src/utils/envVars.ts]: @Battenfield mentioned that y'all discussed on a call this morning that we might want to default this value to altona instead of mainnet? @CarlBeek @hwwhww thoughts?

**Comment by Battenfield** [comment]
> Hey @CarlBeek, when I was QA'ing these changes I noticed that the BLS Verification signature verification includes a check for the upper and lower deposit amounts compared against `depositDatum.amount`, which is always `32*GWEI` (as far as I know).
> 
> I've updated the `MAX_DEPOSIT_AMOUNT` to use `PRICE_PER_VALIDATOR * GWEI`, but now setting `PRICE_PER_VALIDATOR` to anything less than 32 will cause the BLS check to fail, so the user is unable to upload their key file.
> 
> ```
> if (
>     depositDatum.amount < MIN_DEPOSIT_AMOUNT ||
>     depositDatum.amount > PRICE_PER_VALIDATOR * GWEI
>   ) {
>     return false;
>   }
> ```

@CarlBeek and I discussed on call, and decided to keep the original check here. The minimum is fine as inputed and the upper bound would be `32*GWEI`. Hence, this check would no longer be dependent on the `PRICE_PER_VALIDATOR` variable.

File: src/utils/envVars.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (Battenfield) [reply src/utils/envVars.ts:23]: @hwwhww @CarlBeek Can you confirm we will be defaulting to Altona. Also what is the GENESIS_FORK_VERSION for Altona, is it `0x00000000` or another value?

**Comment by Battenfield** [comment]
Few other things that I synced with @georgeweiler on:

- Adding in the GENESIS_FORK_VERSION, defaulting it to `0x00000000`
- Adding in a specific CHAIN_NAME to the environment variables and in the cli instructions
- Making sure the copy for mainnet & testnet are updated and creating pill component for the name in the `<AppBar>` 

very rough mockup of a pill for a testnet or mainnet:
Screen Shot 2020-07-02 at 1 52 22 PM

if on testnet the copy would change to: "... secure the eth2 altona testnet"
Screen Shot 2020-07-02 at 1 50 28 PM


@hwwhww and @CarlBeek let us know if there are any small changes on this. Please take a look at the defaulted altona GENESIS_FORK_VERSION to see if it's correct.

**Comment by georgeweiler** [comment]
@CarlBeek  @Battenfield  @hwwhww I've made some updates to address the comments in this thread. Can I get a double-check on everything? 🙏 

Screen Shot 2020-07-02 at 6 55 17 PM
Screen Shot 2020-07-02 at 6 55 11 PM

**Comment by CarlBeek** [comment]
I like the landing page text for Altona and the text/button on the subsequent pages is exactly what I had in mind. I think it works well to clarify what is happening. Two nitpicks:

1.  The theming on the Altona button doesn't quite fit in with the rest of the page, it feels very suck on (for want of a better word). Maybe it would look better with alpha transparency (see attached pic) or a grey border?

1. Can you add one last env-var for distinguishing between GöETH and ETH please. That way we can swap out the appropriate word whenever we mention the currency.


Screenshot 2020-07-03 at 10 24 36 PM

File: src/pages/Landing/Hero.tsx [hunk 7]
```diff
@@ -75,7 +76,8 @@ export const Hero = () => {
                   color="brand"
                   className="my20"
                 >
-                  Become a validator and help secure the eth2 network.
+                  Become a validator and help secure the eth2{' '}
+                  {IS_MAINNET ? ` mainnet` : ` ${ETH2_NETWORK_NAME} testnet`}.
```
hwwhww, line 80: IMHO we can highlight `Altona testnet` again to warn the users that they are playing on testnet.
Or even note that it's testnet on the header banner, e.g., `Eth2 Launch Pad for Altona testnet`.
image

.env.template
```diff
@@ -0,0 +1,22 @@
...
+# string
+REACT_APP_CONTRACT_ADDRESS=0x16e82D77882A663454Ef92806b7DeCa1D394810f
+# number
+REACT_APP_ETH_REQUIREMENT=524288
+# number
+REACT_APP_PRICE_PER_VALIDATOR=32
+# string
+REACT_APP_ETH2_NETWORK_NAME=Altona
+# Uint8Array such as "0x00000000"
+REACT_APP_GENESIS_FORK_VERSION=0x00000000
```
hwwhww, line 20:
```suggestion
REACT_APP_GENESIS_FORK_VERSION=0x00000121
```

src/utils/envVars.ts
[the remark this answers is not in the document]
CarlBeek: Handled in https://github.com/ethereum/eth2.0-deposit/pull/90#discussion_r449733234

src/pages/Landing/Hero.tsx
CarlBeek: I agree that more clarification is a good idea. I've set up #93 to address this and add more clarification on the matter..

CarlBeek:
I think this is ready for merge. :) As per @hwwhww's earlier comment, it works agains the `dev` branch of the deposit_cli for Altona deposits 🎉  and the remaining aesthetic points are addressed in #93.

Issue #94, opened 2020-07-08
Progress bar after genesis threshold


CarlBeek:
After the genesis deposit threshold is reached the progress bar needs to change to reflect the new state of the network.

Currently:
* The "Launch threshold" is negative
* The label "Launch Threshold" does not match up with the value displayed before the threshold is reached. It should read "XXX ETH to go!" and I think "Already staked" is a more appropriate title for the current staking balance

![Screenshot 2020-07-08 at 5 01 56 PM](https://user-images.githubusercontent.com/12530043/86936986-fd4c7180-c13e-11ea-94ea-c3d44dc5e153.png)

I propose we fix the launch threshold in place and allow the current staking value to overtake it when the balance is larger like in the below image:

![Combined](https://user-images.githubusercontent.com/12530043/86937016-03dae900-c13f-11ea-8a7c-b2dc8fbdc767.png)

Battenfield:
@georgeweiler

CarlBeek:
Closed by #100

PR #100, opened 2020-07-17
Progress bar after Genesis threshold

flamingYawn:
This PR addresses the issue referenced in [this issue](https://github.com/ethereum/eth2.0-deposit/issues/94#issue-653386381).

Although we wanted to add the bar showing how much past the Genesis threshold the current deposited amount is, we realized that the text would overlap under the bar at some points. To remedy this, we temporarily left the bar out and just switched the sides that the text appears on.

Also, different text will show above the bar depending on whether the threshold has been passed and some mobile responsiveness updates were also added.


CarlBeek:
I don't know how to review this. I am not sure how to run the online mode locally, so I can't get this section to display.

flamingYawn:
@CarlBeek 

Hmm.. not sure what you mean by online mode. @Battenfield ?
But in the meantime, here are a few screenshots:

### When threshold has been passed

image

### When threshold is exactly equal with the amount staked
> _Note: This is essentially treated the same as the previous case_

image

### When the threshold has not yet been reached

image

Issue #99, opened 2020-07-17
Insert the Rhino mascot on landing page & as favicon


CarlBeek:

CarlBeek:
Closed via #103

PR #103, opened 2020-07-17
Insert rhino mascot on landing page and as favicon

flamingYawn:
This PR addresses [this issue](https://github.com/ethereum/eth2.0-deposit/issues/99#issue-659536371).


CarlBeek:
Nice job, @flamingYawn. I think there need to me some minor tweaks to the placement of the Rhino. It looks too low on desktop. I think Leslie's face should line up with the header text.
Screenshot 2020-07-18 at 4 43 46 PM

Currently, it centers the image, but it looks funny due to the dust puffs going far to the right of the rhino. Basically I think it should be shifted left a bit. This problem is particularly noticeable on mobile.

Screenshot 2020-07-18 at 4 44 29 PM

flamingYawn:
@CarlBeek I adjusted it a bit.

I couldn't move it to the right very much when not mobile. It starts to overflow the page a bit to the right if I go any further.
My workaround for the mobile case was to add a `transform: scale(1.25, 1.25);` and hide the overflow of the container component. I think that result had mobile looking super nice.

PR #104, opened 2020-07-17
WIP: display beacon explorer links, update network bubble styling

georgeweiler:
Screen Shot 2020-07-17 at 4 30 25 PM

### DIFF
--- a/src/components/AppBar.tsx
+++ b/src/components/AppBar.tsx
@@ -29,7 +29,6 @@ const EthLogo = styled.img`
 
 const NetworkText = styled(Text)`
   margin-right: 30px;
-  background: white;
   padding: 5px 15px;
   border-radius: 4px;
   border: 1px solid;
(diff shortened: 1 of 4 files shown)


georgeweiler:
Hey @CarlBeek do you know if there's a significant delay between when the web3 tx is confirmed and the tx is available on a beacon explorer? I'm seeing a "tx not found" for links that look valid:

https://beaconscan.com/altona/validator/a2038780b47e75470cfe0d428c953f5c8434f9c538c236d0be078d5f6595ae0a17bb0dec997435b3f55c7bb5c1f24e7b#deposits

CarlBeek:
There was a missing `0x` on the pubkeys which beaconscan requires.

Having tested both again, they both work, however, beaconscan takes about a minute to reflect the deposit. (Beaconcha.in does so as soon as the TX is mined (At the point the link becomes availible))

PR #107, opened 2020-07-21
Add client pages

georgeweiler:
Adds the validator client routes for `/nimbus` `/lighthouse` `/prysm` and `/teku`.

Screen Shot 2020-07-21 at 9 47 05 AM
Screen Shot 2020-07-21 at 9 47 17 AM


CarlBeek:
@georgeweiler, can you check my latest commits. Nothing fancy, just fixed some of the client content & added links from the client cards to the client pages

georgeweiler:
if you are happy with the copy, those commits lgtm! 👍

Issue #41, opened 2020-04-09
Differentiate between invalid deposit file and bad deposit data


CarlBeek:
When submitting a `deposit_data.json`, there are two failure modes.  The first is when the user just uploads a random file which is not the correct file. The second is when a user uploads a `deposit_data.json` and the file is not valid/is corrupt/has invalid signature(s). The second state is much more severe and we should display a much more ominous error message encouraging users to get in contact opening an issue here or on `ethereum/eth2.0-deposit-cli`.

sachayves:
Is it expected behaviour that one can submit the same `deposit_data` file twice?

CarlBeek:
No. Using the same deposit data more than once, will result in double-deposits for validators which will just lock up all fund >=32 ETH until transfers are enabled. This is another class of incorrect deposit_data.

Battenfield:
Solved in #108

PR #108, opened 2020-07-21
Differentiate between invalid deposit file and bad deposit data

flamingYawn:
This PR addresses [this issue](https://github.com/ethereum/eth2.0-deposit/issues/41)
image


CarlBeek:
This is part of the way there, but does not quite solve the problem. There are a few classes of failures and we shouldn't have this scary error for all of them

if the fields don't match the intended deposit-data json structure:
    display error about uploading the wrong file
else if the fork version doesn't match the one from the envVar:
    tell the user that this deposit data is not intended for this ${network}
else:
    display this scary message

CarlBeek:
You got these changes in quickly, nice!

Testing it, it now achieves the majority of what I think it should, but there is still one distinction that should be made between a random json file and a deposit_dat.json. The problem now is that if someone upload their keystore instead of the deposit data (highly likely) they see the Big Scary Message (tm) instead of the "this doesn't appear the be the correct file" message.

flamingYawn:
Hmm. Are you sure?

If the structure of the json file doesn’t match what it should be (at least through all of my testing), it should just say “`blah.json` is invalid.”

Tell me what you find... because AFAIK the scary error will only show up if an error is thrown in the `try` (which I haven’t actually seen happen since I made these changes).

src/pages/UploadValidator/index.tsx
```diff
@@ -128,11 +178,23 @@ const _UploadValidatorPage = ({
...
+              ) {
+                // file doesn't match the correct network
+                handleWrongNetwork();
+              }
             }
           } catch (e) {
             // possible error example: json is invalid or empty so it cannot be parsed
             // TODO think about other possible errors here, and consider if we might want to set "isFileStaged"
             setIsFileAccepted(false);
+            handleSevereError();
```
flamingYawn, line 197: To be particular, this is the only time the scary error is shown.

CarlBeek: Indeed, but uploading `keystore.json` instead of `deposit_data.json` when they are both in the same directory is a very easy thing to do and then the user sees the Big Scary.

flamingYawn: I think I’m misunderstanding you since that scenario wouldn’t wouldn’t reach the catch block. My apologies for that (insomnia, maaaaan. haha)

Are you saying that somebody _should_ get the scary error if they try uploading their `keyfile.json`?

flamingYawn: Would it be better to show a different scary message than the BigScary™️ one that’s shown now that conveys something like

> “Dude. This is your keyfile, not the deposit_data. You should be guarding this with your life. Try again.”

?

CarlBeek: > (insomnia, maaaaan. haha)

Can relate. I actually think I checked out the wrong version of this branch. I tried it again and I think it is fine, apologies.

src/pages/UploadValidator/index.tsx
```diff
@@ -102,7 +109,51 @@ const _UploadValidatorPage = ({
...
+          open an issue on GitHub
+        </Link>
+        &nbsp;to get in contact with us.
+      </>
+    );
+  };
+
+  const onFileDrop = (jsonFiles: Array<any>, rejectedFiles: Array<any>) => {
+    if (rejectedFiles?.length) {
+      setFileError(<>Please use a valid json file</>);
```
CarlBeek, line 153:
```suggestion
      setFileError(<>That is not a valid deposit_data json file.</>);
```

PR #125, opened 2020-07-23
Clarifies numbers on network status

CarlBeek:
The text of the Network status graphs is rather unclear as to what numbers are what (see picture below), this PR clarifies that.

Screenshot 2020-07-23 at 4 20 04 PM


CarlBeek:
Merging in favour of #126

PR #154, opened 2020-08-03
Make sure the user is using `master` branch

hwwhww:
Update the context and git clone command to make sure that the user is using the `master` branch.


CarlBeek:
LGTM 👍

Issue #150, opened 2020-07-31
Wrong deposited GoETH in congratulations page


PlanetaOTC:
There is a bug in the launchpad. If you for example want to setup 2 validators but only make the deposit in one of them and later click continue, the next page says that you have deposited 64 GoETH instead only 32. Not sure where or to who report this. Maybe some of the prysm devs know

samajammin:
Thanks for reporting @PlanetaOTC!

Sorry for the long delay here 😅  but this was solved in https://github.com/ethereum/eth2.0-deposit/pull/366

PR #171, opened 2020-09-15
Accurate deposit balances & # of new validators

CarlBeek:
Addresses #150 whereby #of new validators & amount deposited was inaccurate as all deposits in the deposits_data were just assumed to have succeeded. Now, the actual number of deposits are checked based off of the transaction success property for each deposit.


samajammin:
Hey @CarlBeek - I believe this was solved in https://github.com/ethereum/eth2.0-deposit/pull/366

Feel free to re-open if you think I'm missing something. Thanks!

PR #174, opened 2020-09-21
Eth1 Content

CarlBeek:
## Changes
* References to Eth1 "Validators" changed to "clients" as that is what they are
* Adds content for all the Eth1 Clients
* Replaces internal references from "Parity" to "OpenEthereum"


Battenfield (commented):
👍 Looks like mostly copy but there are a few changes to the routes. @georgeweiler can you double-check these minor logic changes?

CarlBeek:
The routes changes should only relate to naming, but indeed please check

flamingYawn:
Looks good to me, @CarlBeek.

One thing though: this PR's [target branch](https://github.com/ethereum/eth2.0-deposit/tree/finish-select-validator) was merged into `add-select-validator-to-workflow` already. Should we update this PR to do the same?

georgeweiler:
@flamingYawn I just changed the target branch and rebased it instead of waiting for @CarlBeek to respond async. LGTM. Thanks @CarlBeek for cleaning up those var names!

PR #176, opened 2020-09-22
Update Hero.tsx (change payouts to rewards)

sachayves:
as per a conversation with danny, the inability to immediately withdraw or transfer means the term payouts may be misleading.


CarlBeek (approved):
Good point, this is indeed more accurate.

Issue #183, opened 2020-09-24
[Urgent] Return of the double deposits


CarlBeek:
There is another bug in the transaction handling process. Basically, it is possible to initiate the same transaction multiple times by clicking start more than once without signing the actual TX. This occurs because we check that the status is `READY` before initiating a transaction, but it is only updated once the user clicks sign. Between these two moments, the user can launch as many instances of the same transaction as they want.

CarlBeek:
Resolved via #184 and merged into master by #185

PR #184, opened 2020-09-24
[Urgent] Update TX status to PENDING in `handleMultipleTransactions`

CarlBeek:
Solves double deposit issue explained in #183. In the change from offering one-at-a-time deposit processing to multiple deposits via the "Initiate All Deposits" button, the status of a transaction was no longer updated to reflect that the TX was `PENDING`. This PR solves that as well as cleans up the `handleMultipleTransactions` recursion.


georgeweiler:
lgtm.  I like the refactor to check the status while recusing instead of relying on the index  👍

georgeweiler (approved):

PR #186, opened 2020-09-25
Update index.tsx

sachayves:
Make it clear that one shouldn't expect multiple keys if one is just running 1 validator


CarlBeek (approved):
LGTM

PR #187, opened 2020-09-25
Rework Generate keys step to leverage tabs

flamingYawn:
This separates the two different options for generating keys into tabs instead of stacking them:
* Downloading binary from `ethereum/eth2.0-deposit-cli` releases
* Building the cli from source

image
image


src/pages/GenerateKeys/Instructions.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <Heading level={2} size="small" color="blueMedium">
+        How do you want to run the CLI?
+      </Heading>
+      <div className="my20" style={{ display: 'flex' }}>
+        <TextSelectionBox
+          isActive={selectedOption === 1}
+          onClick={() => setSelectedOption(1)}
+          style={{ marginRight: '20px' }}
+        >
+          Download pre-built binary
```
CarlBeek, line 26:
```suggestion
          Download CLI app
```

CarlBeek (approved):
This is exactly what I had in mind, thanks @flamingYawn! I think it is much clearer that these are two mutually exclusive options as opposed to one monolithic CLI installation process.

hwwhww (approved):
I love this PR a lot! 👍 

I feel "pre-built binary" is more clear for technical users, and "CLI app" is more understandable for non-technical users. I'm fine with either way. :)

CarlBeek:
Technically, pre-built binary is more correct, but in the past we decided to go with CLI app (eg. #162). I'm going to go with app for now (for consistency) and we can change it down the line if we need to.

PR #190, opened 2020-09-29
Fix Prysm Spadina Documentation Link

rauljordan:
Fixes https://github.com/ethereum/eth2.0-deposit/issues/179, this PR adds the right documentation link for Prysm's Spadina testnet support

Issue #155, opened 2020-08-03
Use envVars configurations on deposit-cli instruction


hwwhww:
#146 hotfixes the instruction. To prepare for mainnet, we should use https://github.com/ethereum/eth2.0-deposit/blob/dev/src/utils/envVars.ts to set `--chain ${CHAIN_NAME.toLowerCase()}` later.

PR #191, opened 2020-10-01
Make installation URLs to ENV vars

hwwhww:
### Issue
Address #155 

Replace #190: We will have to change setting `REACT_APP_PRYSM_INSTALLATION_URL` when we have Prysm's Zinken installation guide.

### DIFF
--- a/src/pages/Clients/Eth2/Lighthouse.tsx
+++ b/src/pages/Clients/Eth2/Lighthouse.tsx
@@ -9,6 +9,7 @@ import {
 } from '../ValidatorClientComponents';
 import { Text } from '../../../components/Text';
 import { Link } from '../../../components/Link';
+import { LIGHTHOUSE_INSTALLATION_URL } from '../../../utils/envVars';
 
 const CodeSnippet = styled.div`
   padding: 10px;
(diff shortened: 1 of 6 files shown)


src/pages/Clients/Eth2/Lighthouse.tsx
```diff
@@ -56,12 +57,8 @@ export const LighthouseDetails = ({ shortened }: { shortened?: boolean }) => (
       <SectionTitle level={2} className="mb5">
         Installation
       </SectionTitle>
-      <Link
-        external
-        primary
-        to="https://lighthouse-book.sigmaprime.io/installation.html"
-      >
-        https://lighthouse-book.sigmaprime.io/installation.html
+      <Link external primary to={LIGHTHOUSE_INSTALLATION_URL} withArrow>
```
hwwhww, line 60: I think the husky linter changes the format automatically when I commited. Do I have to force revert the format?

CarlBeek: Husky is very aggressive, and I'm always nervous when I see all so many files being changed when committing. That said, it does keep the formatting consistent and I've made peace with the git-history having many such changes. (If you force revert now, it will just update the ext time someone commits.)

CarlBeek (approved):
This is a much better way of maintaining the correct links to clients. 👍

CarlBeek:
Note: this must not be merged into master until the Zinken envars have been updated.

PR #192, opened 2020-10-01
Check depost_cli_version of the deposit_data*.json file

hwwhww:
## Issue
See https://github.com/ethereum/eth2.0-deposit-cli/pull/126
To make sure the launchpad and deposit-cli are on the same page, we can check the `deposit_cli_version` of deposit_data*.json file.

## Important
* It REQUIRES https://github.com/ethereum/eth2.0-deposit-cli/pull/126
* We have to change `REACT_APP_MIN_DEPOSIT_CLI_VERSION` to `0.4.0` for Zinken.


yarn.lock
```diff
@@ -4501,7 +4501,7 @@ commondir@^1.0.1:
   resolved "https://registry.yarnpkg.com/commondir/-/commondir-1.0.1.tgz#ddd800da0c66127393cca5950ea968a3aaf1253b"
   integrity sha1-3dgA2gxmEnOTzKWVDqloo6rxJTs=
 
-compare-versions@^3.5.1:
+compare-versions@^3.5.1, compare-versions@^3.6.0:
```
CarlBeek, line 4504:
```suggestion
compare-versions@^3.6.0:
```
We were using 3.6 anyway, no need to support both

hwwhww: It's autogenerated every time I run `yarn`. I tried to remove `compare-versions@^3.5.1,` but it keeps coming back! 😅

CarlBeek: Never mind, it came down to my poor understanding of `yarn.lock` syntax. What is actually happening is `husky` requires `compare-versions "^3.5.1"` and this line basically says if you are looking for `compare-versions@^3.5.1`, or `compare-versions@^3.6.0`, then the following is the package you want.

CarlBeek (approved):

Issue #194, opened 2020-10-01
Explain why you need to choose an eth1 client


sachayves:
It's not intuitively obvious to someone staking on eth2 why they need to run an eth1 client.
Could do with a simple explanation at the top of the Select Client page.
![Screenshot 2020-10-01 at 22 44 00](https://user-images.githubusercontent.com/5483559/94861362-b3baac80-0437-11eb-9452-5837961898fb.png)

sachayves:
Can't seem to self-assign myself, but will try and post something here today

sachayves:
@CarlBeek do you think it's worth mentioning Infura (and explaining the convenience vs decentralisation tradeoff)?

CarlBeek:
I don't think so. If someone is prepared to configure the infura API and to pay for its use, then IMO, they are welcome to.

sachayves:
To make sure i understand, it's not worth us mentioning Infura on this page?

hwwhww:
closing in favor of #197

PR #197, opened 2020-10-05
Explain why you need to choose an eth1 client

sachayves:
addresses https://github.com/ethereum/eth2.0-deposit/issues/194


CarlBeek (approved):
LGTM

PR #200, opened 2020-10-07
Update Teku.tsx

cemozerr:
Update Teku key management options. 

The key management part now looks like:

Screen Shot 2020-10-07 at 3 01 50 PM


hwwhww (approved):
Thanks @cemozerr!

PR #206, opened 2020-10-13
Adds Checklist page

georgeweiler:
Screen Shot 2020-10-13 at 2 19 20 PM

Screen Shot 2020-10-13 at 2 19 29 PM

PR #211, opened 2020-10-14
adds CTA for direct download link to cli tool

georgeweiler:
Screen Shot 2020-10-14 at 3 24 58 PM

### DIFF
--- a/src/pages/GenerateKeys/Option1.tsx
+++ b/src/pages/GenerateKeys/Option1.tsx
@@ -52,7 +52,7 @@ export const Option1 = ({
           external
           to="https://github.com/ethereum/eth2.0-deposit-cli/releases/"
         >
-          here
+          ethereum/eth2.0-deposit-cli
         </Link>
         .
       </Text>
--- a/src/pages/GenerateKeys/Option1.tsx
+++ b/src/pages/GenerateKeys/Option1.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components';
 import { Heading } from '../../components/Heading';
 import { Text } from '../../components/Text';
 import { Link } from '../../components/Link';
+import { Button } from '../../components/Button';
 import { Alert } from '../../components/Alert';
 import { CHAIN_NAME, IS_MAINNET } from '../../utils/envVars';
 import { colors } from '../../styles/styledComponentsTheme';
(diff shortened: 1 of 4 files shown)


src/pages/GenerateKeys/Option1.tsx
```diff
@@ -18,32 +19,50 @@ export const Option1 = ({
...
+          inline
           primary
           external
           to="https://github.com/ethereum/eth2.0-deposit-cli/releases/"
-          inline
         >
-          ethereum/eth2.0-deposit-cli
-        </Link>{' '}
-        GitHub page for your platform.{' '}
+          here
```
hwwhww, line 55: IMHO it's better to write down the full name for users to verify it.

```suggestion
          ethereum/eth2.0-deposit-cli
```

src/pages/GenerateKeys/Option1.tsx
```diff
@@ -18,32 +19,50 @@ export const Option1 = ({
...
-        </Text>
-      </Alert>
-      <Text weight={500}>
+      <Link external to={cliLink} style={{ margin: '2rem auto' }}>
+        <Button
+          primary
+          label={<Text color="white">Download the CLI tool for {osName}</Text>}
+        />
+      </Link>
+      <Text weight={500} className="mt10">
```
hwwhww, line 65: We have concerns if this pattern is "too convenient" for users so that they won't verify if the source is legit. If we want to add a direct link, IMHO then we *have* to include the checksum (#206) together.

Battenfield: I agree @hwwhww. I don't want to merge this unless it's done in a verifiable way. We could skip this PR in this release until we feel comfortable that it's a safe option. I do think that this will reduce confusion if we can do this properly.

hwwhww: @CarlBeek's checksum PR on the way: https://github.com/ethereum/eth2.0-deposit-cli/pull/133

CarlBeek: Checksum PR helps, not sure whether it solve the verification problem. People could just not verify the checksum. That said, most other apps ([eg geth](https://geth.ethereum.org/downloads/)) offer hosted binaries, so I guess it is ok

CarlBeek: EnvVars for checksums?

hwwhww: I suggest also put download URLs to EnvVars: https://github.com/ethereum/eth2.0-deposit/pull/222
And also check the version.

samajammin:
@hwwhww @CarlBeek do you believe this PR is still relevant? Or should we close this out?

PrapatsaraNaksuth:
@@hwwhww @CarlBeek 
I agree that Checksum PR should be very helpful and convenient for users with initial experience, but there should be more binaries and URL Download. for monitoring. (This is my personal opinion) because I am a yes person who has no basic experience.

PrapatsaraNaksuth (commented):
@@hwwhww @CarlBeek
I agree that Checksum PR should be very helpful and convenient for users with initial experience, but there should be more binaries and URL Download. for monitoring. (This is my personal opinion) because I am a yes person who has no basic experience

Grzegorzkoc (changes requested):
Ggggghh

wackerow:
@CarlBeek Pinging on this... Seems to have stagnated, is this something we're still considering adding?

wackerow:
Going to close this out as it's gone stale. If anyone would like to resurface, please feel free to re-open and post more information.

PR #216, opened 2020-10-15
Remove client cards from congrats page

Battenfield:
Eth2 Client pages are now reachable from the Navbar and the client cards are on the Checklist page.
Screen Shot 2020-10-14 at 11 27 10 PM


hwwhww (approved):
LGTM

PR #218, opened 2020-10-15
Remove reference to phase "3"

### LINKED ISSUE #169: It says "Phase 0, 1, 2 and 3 duration " on the frontpage
On the frontpage of the deposit launchpad a link text says:

Please correct me if I am wrong but so far I havent heard of a phase 3.

CarlBeek:
Closes #169

### DIFF
--- a/src/pages/Landing/Phases/index.tsx
+++ b/src/pages/Landing/Phases/index.tsx
@@ -55,7 +55,7 @@ const phases: phase[] = [
       'The duration for the deployment of the individual phases depends on many factors.',
     linkUrl:
       'https://docs.ethhub.io/ethereum-roadmap/ethereum-2.0/eth-2.0-phases/',
-    link: 'Phase 0, 1, 2 and 3 duration',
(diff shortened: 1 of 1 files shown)


src/pages/Landing/Phases/index.tsx
```diff
@@ -55,7 +55,7 @@ const phases: phase[] = [
       'The duration for the deployment of the individual phases depends on many factors.',
     linkUrl:
       'https://docs.ethhub.io/ethereum-roadmap/ethereum-2.0/eth-2.0-phases/',
```
hwwhww, line 57: Should we use https://ethereum.org/en/eth2/#roadmap, which includes Phase 1.5?
/cc @djrtwo @jjmstark

CarlBeek: I agree with this idea. I've updated all of the phase descriptions to better explain phase 1.5 and the link now points to the eth2 roadmap instead of the ethhub docs.

hwwhww (approved):
LGTM

PR #220, opened 2020-10-15
Minor style adjustment

hwwhww:
Post merge QA

- [x] 1. Fix client intro style
before:
image

after:
image

- [x] 2. Add `underline` to navBar hover
e.g., 
image

- [x] 3. Alphabetical order clients in navbar

### DIFF
--- a/src/components/AppBar.tsx
+++ b/src/components/AppBar.tsx
@@ -102,10 +102,10 @@ export const AppBar = () => {
           dropAlign={{ top: 'bottom', right: 'right' }}
           dropContent={
             <Box pad="small">
-              <DropdownLink to={routesEnum.prysm}>Prysm</DropdownLink>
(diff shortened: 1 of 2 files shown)


georgeweiler:
Thanks @hwwhww! I had some concerns about the cards on smaller screen sizes, which is why I went with the wrapped design.  

I was just playing around with this solution:

Screen Shot 2020-10-15 at 6 21 12 AM

What do you think? Happy to go with whatever you prefer :)

Battenfield (approved):
👍 LGTM. We can spruce up the navbar in a follow-up PR

Battenfield:
Also, we may want to adjust the cards in a slightly different way as I think they will overlap the borders on a smaller screensize

hwwhww:
@georgeweiler 
your solution looks great! 👍 
To be clear, my front-end skill is miserable. Looking forward to your follow-up PR. :)

PR #221, opened 2020-10-15
Chain flag clean-up

hwwhww:
- [x] 1. Add `--chain mainnet` flag to the Mainnet instruction
    - Reason: Although `mainnet` is the default value in deposit-cli, letting the users copy this instruction can save their time.
- [x] 2. Replace `CHAIN_NAME` with `ETH2_NETWORK_NAME`.
    - It turned out they are always the same value in our context.
- [x] 3. Copy the chain-flag-warning from Option 2 to Option1
image


CarlBeek (approved):
A worthwhile simplification 👍

PR #246, opened 2020-11-04
Makes 1 call to beaconchain API for all validators

CarlBeek:
We've been hitting rate limits (10/minute) on the beaconcha.in API because we are making 1 API call per validator. This PR makes use of the API's ability to respond to multiple simultaneous requests so that only 1 request is made to the API.


georgeweiler:
@CarlBeek I've QA'd this extensively and everything checks out. Great job & thanks!

hwwhww (approved):
Add @georgeweiler's approval 😄

PR #253, opened 2020-11-06
Add minimum env setting helper for devs to run it locally

hwwhww:
Except for the private vars, after #252 these 6 env vars in files are the *minimum* config for running a testnet launchpad locally. I used these to switch between mainnet and testnets by `source .env.<network_name>`.

@Battenfield @georgeweiler @CarlBeek do you think it will be helpful for you? If so, we can include these files to the source code.


CarlBeek (approved):
I do think this is a worthwhile addition as a starting point for testing multiple configs.  I also think it is useful for users who may wish to spin up their own deployment to deposit into a deprecated testnet's contract.Personally, I add an Infura API key to my config, so I'd still have to edit it, but this way I can copy a pre-made config.

PR #268, opened 2020-11-18
Fix minimum geth fast sync disk size

hwwhww:
### Issue
The "at least 140 GB SSD to run geth fast" may be outdated.
Changing it to 290 GB.

---

Thank @nonsense for [finding this issue](https://discord.com/channels/595666850260713488/598292067260825641/778352824114217003).
> A new node takes around 290 GB... and only grows from there on

Also thank @michaelsproul for [confirming the number](https://discord.com/channels/595666850260713488/598292067260825641/778472799324602378):
> I recently fast synced Geth from scratch and it's using 287 GB (122 GB hot + 165 GB freezer)

### DIFF
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -126,7 +126,7 @@ export const Checklist = () => {
             <ul>
               <li>
                 <Text>
-                  For example, you need at least 140 GB SSD to run geth fast
+                  For example, you need at least 290 GB SSD to run geth fast
                   sync on <i>mainnet</i>.
                 </Text>
               </li>
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -126,8 +126,12 @@ export const Checklist = () => {
             <ul>
               <li>
                 <Text>
-                  For example, you need at least 290 GB SSD to run geth fast
-                  sync on <i>mainnet</i>.
+                  For example, you need at least ~300 GB SSD to run geth fast
+                  sync on <i>mainnet</i> in November 2020, assuming you move the{' '}
+                  <i>ancient</i> directory on a different drive. Have in mind
+                  that geth <i>chaindata</i> grows with ~1 GB per day, so you
+                  should account for enough space on your SSD and HDD until you
+                  run maintenance on the node.
                 </Text>
               </li>
             </ul>


src/pages/Checklist/index.tsx
```diff
@@ -126,7 +126,7 @@ export const Checklist = () => {
             <ul>
               <li>
                 <Text>
-                  For example, you need at least 140 GB SSD to run geth fast
+                  For example, you need at least 290 GB SSD to run geth fast
                   sync on <i>mainnet</i>.
```
nonsense, line 130:
```suggestion
                  For example, you need at least 290 GB SSD to run geth fast sync on <i>mainnet</i>, assuming you move the `ancient` directory on a different drive. Have in mind that geth `chaindata` grows with 6 GB per day, so you should account for enough space on your SSD and HDD until you run maintenance on the node.
```

A new node takes ~125GB when synced on day 0, and then continues to grow from there. This is why I initially felt that 140GB is misleading, as the idea of an ETH1 node when running and ETH2 validator is not to just sync the node and shut it down, but rather keep it as a long running service, ideally not updating it every day.

If we assume that a developer starts a Geth node with 290 GB SSD, and they move the `ancient` directory elsewhere, then they will run out of disk space in about a month, at which point they will have to somehow prune the chaindata (not sure how that is done, I usually just start a new node, and deprecate the old one).

nonsense: TODO:
- [ ] Confirm that the `chaindata` directory does indeed grow with 6 GB per day. I took that number based on conversations in Discord, and haven't checked myself.

hwwhww: @nonsense
I chatted with @rjl493456442 from Geth team, the ~6GB/day increment should be for the archive node. 😄

nonsense: @hwwhww that sounds right :) I am pretty sure `chaindata` doesn't grow THAT much per day after a fast sync :) Will update the value in a few days when I have more data, unless someone has a better estimate.

hwwhww: A recent geth benchmark data shared by @rjl493456442:
(not sure which sync method they used at the beginning, so we just look at the the total growth rate)

### The leveldb data growth:
- Need SSD storage
- 1.8 GB/week
- 1351605759047_ pic_hd


### The `ancient` (freezer) data growth:
- `ancient` is okay to be stored with HDD
- 6 GB/week
- 1391605759145_ pic_hd

---

Maybe we can say

```suggestion
                  For example, you need at least ~300 GB HDD to run geth fast sync on <i>mainnet</i> in November 2020,
                  assuming you move the <i>ancient</i> directory on a different drive.
                  Have in mind that geth <i>chaindata</i> grows with ~1 GB per day, so you should account for enough space on your SSD and HDD until you run maintenance on the node.
```

@nonsense What do you think? 🙂

nonsense: @hwwhww perfect! I think this is much better than the current version on master 👍

CarlBeek (approved):
LGTM, thanks @hwwhww & @nonsense

PR #272, opened 2020-11-23
updated for issue with time sync

PatrickAlphaC:
When running through the checklist page, and running the `sudo timedatectl set-ntp on` command, I was running into `Failed to set ntp: NTP not supported`

However installing `chrony` fixed this. `sudo apt install chrony`

So I think it makes sense to either note that something else they will need to install, or just to plop this package in there as well. I was told this way was a bit `overkill`, but hoping this helps users moving forward. Looks like I wasn't the only one tripped up here.


hwwhww:
Thank you @PatrickAlphaC! I cherry-pick-ed your commit and fix CSS in #275 together.
Closing this PR now.

Issue #270, opened 2020-11-20
A strong warning about running validators in two places


superphiz:
Many users will not be aware that running their validator keys in two places will lead to a slashable event and ejection from the network. Users are likely to create redundant systems for backup and may run their keys on two machines to ensure high availability. The launchpad should include a very strong and visible warning that running the same validator keys in two locations will lead to slashing.

Reference: https://www.reddit.com/r/ethstaker/comments/jxro3n/slashing_causes_and_consequences_phase_0/

This is a very well-known event in the intermediate & advanced user communities, but it is likely to be unknown to beginner stakers.

superphiz:
https://www.reddit.com/r/ethstaker/comments/k0q2ic

djrtwo:
Looks like this made it into the checklist

Issue #254, opened 2020-11-06
Add EthStaker to Launchpad


Buttaa:
Hi! 

I have seen a lot of new (confused) users in the client discord servers. The EthStaker community would like to help!
If possible, please add a link to our discord (invite.gg/ethstaker) and reddit (reddit.com/r/ethstaker)

djrtwo:
addressed in #275

PR #275, opened 2020-11-26
Update checklist

hwwhww:
- Add simulation tips from https://github.com/eth2-educators/launch_checklist/blob/main/README.md
- Address #254 
- Address #270
- Address #272 + fix CSS
- Remove "fast" from "Geth fast sync" as @superphiz's suggestion
- @CarlBeek's suggestions
    - Add Eth1 HTTP API check
    - Add clock time check
- Update Teku discord URL
- Add a warning that your Eth1 node may take days to sync.

### DIFF
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -264,8 +264,8 @@ export const Checklist = () => {
             <li className="py5">
               <Text>
                 <strong>Note:</strong> the RTC (Real-Time Clock) time may be set
-                to UTC time instead of your local timezone especially in a VM
-                which has clock configured on Windows.
+                to your local timezone instead of UTC, especially in a VM which
+                has clock configured on Windows.
               </Text>
             </li>
           </ul>
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -243,6 +243,7 @@ export const Checklist = () => {
                 <pre className="my0">
                   <span style={{ color: colors.red.medium }}>
                     {' '}
+                    sudo apt install chrony
                     sudo timedatectl set-ntp on
                   </span>
                 </pre>
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -253,6 +253,22 @@ export const Checklist = () => {
               </Text>
             }
           />
+          <CheckBox
+            label={
+              <Text className="checkbox-label" style={{ display: 'inherit' }}>
+                I've verified my server's time matches the wall clock.
+              </Text>
+            }
+          />
+          <ul className="sub-checklist-item">
+            <li className="py5">
+              <Text>
+                <strong>Note:</strong> time can be set to UTC instead of
+                timezone especially in a VM which has clock configured for
+                windows mode.
+              </Text>
+            </li>
+          </ul>
         </section>
         <section>
           <Heading level={3}>Eth1 Client</Heading>
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -143,37 +143,6 @@ export const Checklist = () => {
           </ul>
         </section>
         <section>
-          <Heading level={3}>Keystore</Heading>
-          <CheckBox
-            label={
-              <Text className="checkbox-label">
-                I have securely generated and saved my keystore(s).
-              </Text>
-            }
-          />
-          <CheckBox
-            label={
-              <Text className="checkbox-label">
-                I have backed up my mnemonic.
-              </Text>
-            }
-          />
-          <CheckBox
-            label={
-              <Text className="checkbox-label">
-                I have set a strong password for my keystore(s)
-              </Text>
-            }
-          />
-          <CheckBox
-            label={
-              <Text className="checkbox-label">
-                I have backed up my password.
-              </Text>
-            }
-          />
-        </section>
-        <section>
           <Heading level={3}>Security</Heading>
           <CheckBox
             label={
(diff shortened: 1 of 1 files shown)


src/pages/Checklist/index.tsx
```diff
@@ -236,14 +241,34 @@ export const Checklist = () => {
               <Text className="checkbox-label" style={{ display: 'inherit' }}>
                 For Ubuntu 20.04:
                 <pre className="my0">
+                  {'\n'}
+                  <span style={{ color: colors.red.medium }}>
+                    sudo apt install chrony
```
hwwhww, line 246: @PatrickAlphaC  I have second thought: the context is about Ubuntu 20.04, so with the default `timesyncd` service it should be fine to sync the time.

src/pages/Checklist/index.tsx
```diff
@@ -123,18 +124,22 @@ export const Checklist = () => {
...
+            </li>
+            <li className="py5">
+              <Text>
+                <strong>Note:</strong> it may take day(s) to sync the Eth1
+                client in the worst-case scenario.
+              </Text>
+            </li>
           </ul>
         </section>
         <section>
```
CarlBeek, line 145: I think we should purge this keystore section. There is not much a user can do if that have performed these checklist items, if they have made deposits, then they don't really matter.

src/pages/Checklist/index.tsx
```diff
@@ -234,16 +239,42 @@ export const Checklist = () => {
           <CheckBox
             label={
               <Text className="checkbox-label" style={{ display: 'inherit' }}>
-                For Ubuntu 20.04:
+                For Ubuntu 20.04, run
                 <pre className="my0">
+                  {' '}
                   <span style={{ color: colors.red.medium }}>
-                    {' '}
                     sudo timedatectl set-ntp on
```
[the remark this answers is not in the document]
CarlBeek: It may also be worth noting that by default VMs may disable NTP and may need some effort on the users part to find a work-around.

src/pages/Checklist/index.tsx
```diff
@@ -234,16 +239,42 @@ export const Checklist = () => {
...
+              <Text className="checkbox-label" style={{ display: 'inherit' }}>
+                I've verified my server's time matches the wall clock.
               </Text>
             }
           />
+          <ul className="sub-checklist-item">
+            <li className="py5">
+              <Text>
+                <strong>Note:</strong> time can be set to UTC instead of
+                timezone especially in a VM which has clock configured for
```
CarlBeek, line 273: What I meant here is that Windows, amongst a bunch of other crazy decisions w.r.t. timezones & timing, sets the RTC to local time and the calculates UTC from there when it is needed. This means that VMs often have an option to have the RTC be in UTC or local time and is some thing to heed. (This is why the `RTC in local TZ` field exists in the `timedatectl` params described above.

src/pages/Checklist/index.tsx
```diff
@@ -341,7 +398,7 @@ export const Checklist = () => {
                 Prysm
               </Link>{' '}
               |{' '}
-              <Link primary inline external to="Teku">
+              <Link primary inline external to="https://discord.gg/7hPv2T6">
                 Teku
               </Link>
             </li>
```
CarlBeek, line 404: Line 409 below is a duplicate reminder to open the ports on the router.

src/pages/Checklist/index.tsx
```diff
@@ -499,6 +556,73 @@ export const Checklist = () => {
...
+                <strong>Hint:</strong> you can recover it with{' '}
+                <Link
+                  inline
+                  external
+                  to="https://github.com/ethereum/eth2.0-deposit-cli#existing-mnemonic-arguments"
+                >
+                  deposit-cli existing-mnemonic command
+                </Link>
+              </Text>
+              .
```
CarlBeek, line 581: Why is this fullstop on a new line?

src/pages/Checklist/index.tsx
```diff
@@ -499,6 +556,73 @@ export const Checklist = () => {
             </li>
           </ul>
         </section>
+        <section>
+          <Heading level={3}>Simulations</Heading>
+          <CheckBox
```
CarlBeek, line 561: I think we should have a description here about what simulations are and why you might want to run them something along the lines of:

```text
The following are are recommended simulations for ensuring your validation setup is robust:
```

src/pages/Checklist/index.tsx
```diff
@@ -499,6 +556,73 @@ export const Checklist = () => {
...
+            label={
+              <Text className="checkbox-label">
+                I have simulated how to recorver my keystore JSON files with
```
CarlBeek, line 564: Again not sure this is useful for someone who has already made the deposits.

src/pages/Checklist/index.tsx
```diff
@@ -499,6 +556,73 @@ export const Checklist = () => {
...
+              <Text className="checkbox-label">
+                I have simulated how to migrate from one Eth2 client to another
+                Eth2 client.
+              </Text>
+            }
+          />
+          <CheckBox
+            label={
+              <Text className="checkbox-label">
+                I have simulated how to change my keystore passphrase and resume
```
CarlBeek, line 603: Is this something we expect users to know how to do? Surely it is easier to just delete and recreate your keystores?

src/pages/Checklist/index.tsx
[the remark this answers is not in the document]
hwwhww: good point! I updated and reorganized the sentences.

src/pages/Checklist/index.tsx
hwwhww: Right, I changed it to "I have simulated how to manually stop and restart my Beacon Node (BN) and Validator Client (VC) gracefully." now.

src/pages/Checklist/index.tsx
hwwhww: Thanks for the clarification!

I updated this note to

"Note: the RTC (Real-Time Clock) time may be set to UTC time instead of your local timezone especially in a VM which has clock configured on Windows."

Does that make sense to you?

src/pages/Checklist/index.tsx
CarlBeek: Sounds good to me:)

src/pages/Checklist/index.tsx
CarlBeek: Yes, but it is the other way round, on windows, the RTC is in local, and on all sane OSs, it is in UTC.

CarlBeek (approved):
These are some helpful additions. Thanks to all who contributed!

src/pages/Checklist/index.tsx
PatrickAlphaC: How can we check that is working? I used a new Ubuntu install and was not able to `set-ntp`

hwwhww: Is your Ubuntu version 20.04?

You can run this command to see the status of the Ubuntu default `systemd-timesyncd` NTP client:
```sh
sudo systemctl status systemd-timesyncd.service
```

If it's inactive, you can (1) try to restart it:
```sh
sudo systemctl restart systemd-timesyncd.service
```

or (2) install the alternative `chrony` or `ntp` pacakges.

PatrickAlphaC: `Failed to restart systemd-timesyncd.service: Unit systemd-timesyncd.service is masked.`

So I did (2) hence why I made this issue / PR

PR #278, opened 2020-11-27
Adds depoit offset amount option for "free" deposits on pyrmont

CarlBeek:
In order to save on GoETH for the Pyrmont testnet, that specific deposit contract was modified to allow early deposits to be made for free. The launchpad calculates the deposit progress based on the balance of the contract and because of the free deposits, the progress bar is wrong. This PR addresses this by adding a balance-offset parameter which is added to the amount when the contract is queried.


src/utils/envVars.ts
```diff
@@ -25,6 +25,12 @@ if(process.env.REACT_APP_ETH_REQUIREMENT && Number.isNaN(Number(process.env.REAC
 }
 export const ETH_REQUIREMENT            = process.env.REACT_APP_ETH_REQUIREMENT     || 524288;
 
+// ETH_DEPOSIT_OFFSET is added to the balance of the deposit contract to account for testnet deposit-contracts that allow some number of free deposit
+if(process.env.REACT_APP_ETH_DEPOSIT_OFFSET && Number.isNaN(Number(process.env.REACT_APP_ETH_DEPOSIT_OFFSET))) {
+    throw new Error("REACT_APP_ETH_DEPOSIT_OFFSET must be of type: number")
+}
+export const ETH_DEPOSIT_OFFSET = Number(process.env.REACT_APP_ETH_DEPOSIT_OFFSET) * Number(IS_MAINNET) || 0;
```
hwwhww, line 32: Why multiply it with `IS_MAINNET`? Do you mean:
```suggestion
export const ETH_DEPOSIT_OFFSET = Number(process.env.REACT_APP_ETH_DEPOSIT_OFFSET) * Number(!IS_MAINNET) || 0;
```
?

(add a `!`)

hwwhww (approved):
1. The fix looks good except for the missing inverse `!`.
2. Also need to update **all** the `.env.<network>` files to 100% overwrite other files.

👍

Issue #279, opened 2020-12-01
"Try again" button doesn't work after failed transaction


dankrad:
After a failed transaction, the "try again" button appears but doesn't do anything.

Unfortunately, a transaction can "fail" from the perspective of the deposit UI for entirely trivial reasons, such as timing out or declining confirmation from the hardware wallet.

PR #280, opened 2020-12-01
Allow retry on failed deposits

CarlBeek:
If a user takes too long to sign a deposit on their HW wallet, or they accidentally reject it on their HW wallet, Metamask returns a different error, this handles those cases. Addresses #279

### DIFF
--- a/src/pages/Transactions/Keylist/ActionButton.tsx
+++ b/src/pages/Transactions/Keylist/ActionButton.tsx
@@ -97,10 +97,7 @@ export const ActionButton = ({
     );
   }
 
-  if (
-    transactionStatus === TransactionStatus.FAILED ||
-    transactionStatus === TransactionStatus.REJECTED
-  ) {
+  if (transactionStatus === TransactionStatus.REJECTED) {
     return (
       <Container onClick={onClick}>
         <ButtonText>Try again</ButtonText>
--- a/src/pages/Transactions/transactionUtils.tsx
+++ b/src/pages/Transactions/transactionUtils.tsx
@@ -121,7 +121,6 @@ export const handleMultipleTransactions = async (
       }
     )
     .on('error', (error: any) => {
-      console.log(error);
       if (isUserRejectionError(error)) {
         updateTransactionStatus(pubkey, TransactionStatus.REJECTED);
       } else {
--- a/src/pages/Transactions/transactionUtils.tsx
+++ b/src/pages/Transactions/transactionUtils.tsx
@@ -58,6 +58,7 @@ export const handleMultipleTransactions = async (
   const remainingTxs = depositKeys.filter(
     key =>
       key.transactionStatus === TransactionStatus.READY ||
+      key.transactionStatus === TransactionStatus.FAILED || // can be triggered by user taking too long to confirm on hardware wallet
       key.transactionStatus === TransactionStatus.REJECTED
   );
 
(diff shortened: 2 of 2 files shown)


src/pages/Transactions/Keylist/ActionButton.tsx
```diff
@@ -97,10 +97,7 @@ export const ActionButton = ({
     );
   }
 
-  if (
-    transactionStatus === TransactionStatus.FAILED ||
```
CarlBeek, line 101: We don't allow transactions to be processes if they are `FAILED`, so there shouldn't be a retry button.

src/pages/Transactions/transactionUtils.tsx
```diff
@@ -13,12 +13,22 @@ const pricePerValidator = new BigNumber(PRICE_PER_VALIDATOR);
...
-    return true; // metamask
+    return true; // metamask reject
+  if (
+    error.message ===
+    '​Error: TransportStatusError: Ledger device: Condition of use not satisfied (denied by the user?) (0x6985)'
+  )
+    return true; // ledger reject via metamask
+  if (
+    error.message ===
+    '​Error: TransportError: Failed to sign with Ledger device: U2F OTHER_ERROR'
```
CarlBeek, line 29: This can be reached in at least two ways:
* The user can have no Ledger connected or not click `confirm` in time, in this case I'd like to return `true` so the TX can be retried.
* The user can time out on the ledger device, but later click confirm. In this case, the TX will still go through.

Unfortunately Metamask doesn't distinguish between these cases (in error message or number), and therefore it is unsafe to offer the option to retry in case the user clicks `Accept` after Metamask times out and the TX is submitted (opening up the risk of a double deposit).

CarlBeek: I've tried with Metamask in English, Dutch, and German (the latter two were clearly google-translated), the errors are always the same. On the ledger side, I was surprised to learn that they only support English for both their mnemonics, and the firmware running on their devices (Ledger Live can run in other languages though).

It is enough to distinguish the cases and making this change actually helped me catch a bug with an invisible unicode character.

src/pages/Transactions/transactionUtils.tsx
```diff
@@ -13,16 +13,28 @@ const pricePerValidator = new BigNumber(PRICE_PER_VALIDATOR);
...
-    error.message ===
-    'MetaMask Tx Signature: User denied transaction signature.'
+    error.message.includes(
+      'Ledger device: Condition of use not satisfied (denied by the user?)'
+    ) &&
+    error.code === -32603
   )
-    return true; // metamask
-  if (error.message === 'User denied transaction signature.') return true; // portis
+    return true; // ledger reject via metamask
```
hwwhww, line 31: It looks like it could be refactored the `error.code === -32603` a bit
```js
  if (error.code === -32603)
    // do nothing because there are different cases with the same error-code ;
    if (
        // (1) metamask reject
        error.message.includes(
        'MetaMask Tx Signature: User denied transaction signature.'
        ) || (
        // (2) Ledger reject via metamask
        error.message.includes(
        'Ledger device: Condition of use not satisfied (denied by the user?)'
        ))) {
        return true;
    }
```

Does `Fortmatic RPC Error: [-32603] Fortmatic: User denied transaction.` case also has `error.code === -32603`?

CarlBeek:
Finally managed to neaten this up. Hardware wallet error reporting via Metamask is very bad. Error messages are undocumented, differ between wallets and different failures can have the same error message.

hwwhww (approved):
LGTM
Great work @CarlBeek

PR #290, opened 2020-12-16
Makes deposit command compatible with all shells.

MicahZoltu:
Some shells (notably Powershell on Windows) require a leading `.\` when executing something from the current directory.  This also ensures that if the user has some globally installed `deposit` binary on their machine that won't be used and only the `deposit` file in the current directory will be or the command will fail (a good thing in this case).


hwwhww (approved):
LGTM! Thanks @MicahZoltu!

PR #291, opened 2020-12-18
WIP: top up feature

georgeweiler:


CarlBeek:
This looks to be a good start in this direction, but I've encountered a few bugs testing this out:
- [ ] If the validator has deposited, but has bot been processed by the beaconcha.in queue, it technically doesn't exist in eth2 yet and therefore the second request (to the the effective balance) doesn't return anything for that address which causes the page to crash. Instead if nothing is returned for a given address, we should not allow more deposits.
- [ ] If there is only 1 deposit for a given eth1 address, then beaconcha.in doesn't return a length 1 array, but returns the object inside. This also causes the page to crash.

CarlBeek (commented):
I like how this works. The spinners in the TX processing are a nice touch. 👍

CarlBeek:
As per our discussion the other day, let's get this merged! 💪

I'm happy with how it works now save two items:
* The linting error about React Hook `React.useMemo` missing a dependency (`showDepositVerificationWarning`): ∫https://github.com/ethereum/eth2.0-deposit/pull/291/files#diff-f810da4dbf04ce49a49c3ae5db5acff4dee8737c515068c44945b393fbceac58R126
* re-enable the validity checks on the submit button: https://github.com/ethereum/eth2.0-deposit/pull/291#discussion_r555918723

package.json
```diff
@@ -10,6 +10,7 @@
     "@types/numeral": "^0.0.28",
     "@types/react-animate-on-scroll": "^2.1.2",
     "@types/react-copy-to-clipboard": "^4.3.0",
+    "@types/react-js-pagination": "^3.0.3",
```
samajammin, line 13: Is there any need for this package? I don't see it used anywhere.

package.json
```diff
@@ -30,13 +31,15 @@
     "grommet-icons": "^4.4.0",
     "history": "^4.10.1",
     "lodash": "^4.17.15",
+    "moment": "^2.29.1",
```
samajammin, line 34: Same here.

package.json
```diff
@@ -30,13 +31,15 @@
...
     "numeral": "^2.0.6",
     "react": "^16.12.0",
     "react-animate-on-scroll": "^2.1.5",
     "react-animated-css": "^1.2.1",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.12.0",
     "react-dropzone": "^10.2.1",
+    "react-js-pagination": "^3.0.3",
```
samajammin, line 42: Same here.

samajammin:
@CarlBeek I've made the changes you requested & removed a few unnecessary packages from this PR. I believe it's ready to go - please merge if you agree! Thanks.

CarlBeek (approved):
Thanks for cleaning this up @samajammin and @georgeweiler for implementing!

package.json
CarlBeek: I can't see any good reason, happy you removed. 👍

PR #297, opened 2021-01-11
Removed duplicated word typo

wackerow:
Small typo correction on the checklist page (`are are`): 
image
Simple correction.


CarlBeek (approved):
Thanks @wackerow, so easy to miss these!

PR #299, opened 2021-01-15
FAQ: Modify initial validator ETH balance question

### LINKED ISSUE #298: Unclear statements in ETH 2 launchpad FAQ page
Hey there,

I was reading the FAQ in https://launchpad.ethereum.org/faq and these two particular answers stood out to me:

> **How much ETH does a validator need to stake?**
> Before a validator can start to secure the network, he or she needs to stake 32 ETH. This forms the validator’s initial balance.

> **Is there any advantage to having more than 32 ETH at stake?**
> No. There is no advantage to having more than 32 ETH staked.
> Limiting the maximum stake to 32 ETH encourages decentralization of power as it prevents any single validator from having an excessively large vote on the state of the chain.
> Remember that a validator’s vote is weighted by the amount it has at stake.

On my first read, I was confused as to how a validator's vote is weighted by the amount it has at stake. I understood that both the minimum and maximum ETH required to stake is 32, whereas it's a bit more nuanced.

I think that this would be clearer if we mention the following: 32 ETH is the minimum needed to initially active the validator. You can eventually stake less than that (e.g if you withdraw in the future, or penalties lower you below 32) but to **start** staking you need 32.

stanislavkozlovski:
Fixes #298

### DIFF
--- a/src/pages/FAQ/index.jsx
+++ b/src/pages/FAQ/index.jsx
@@ -90,8 +90,7 @@ export const FAQ = () => {
             </Heading>
             <Text className="mt10">
               <strong>32 ETH</strong> is the amount needed to initially active the validator - this forms the validator's initial balance which itself can shrink or grow.
-              A validator can eventually stake less than that (e.g if penalties lower the balance) down to 16 ETH.
-              It cannot stake more than 32 ETH.
+              A validator can eventually stake less than that (e.g if penalties lower the balance), down to 16 ETH.
             </Text>
           </section>
           <section>
--- a/src/pages/FAQ/index.jsx
+++ b/src/pages/FAQ/index.jsx
@@ -90,7 +90,7 @@ export const FAQ = () => {
             </Heading>
             <Text className="mt10">
               <strong>32 ETH</strong> is the amount needed to initially active the validator - this forms the validator's initial balance which itself can shrink or grow.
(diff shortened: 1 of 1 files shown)


stanislavkozlovski:
cc @georgeweiler @CarlBeek @sachayves let me know if this makes sense to you

CarlBeek:
Penalties (including leaking) lowering the balance are the only way you could stake <32 ETH. I think suggesting that you can stake 32 is also confusing as to me it implies it so something you'd want to do.

stanislavkozlovski:
Do you mean that "suggesting that you can stake less than 32 is also confusing" ?

samajammin:
Thanks for the PR @stanislavkozlovski!

@hwwhww @CarlBeek do you believe this PR is still relevant? Or should we close this out?

hwwhww:
Given that this question was changed to a more accurate question: "How much ETH do I need to stake to become a validator?", I think it's okay to close it now. Thank you @stanislavkozlovski!

jokerksa12:
I am not sure much, but I see the shutdown is safe at this time. Thank you
for your efforts and I apologize for not responding because I am new to
this program and I do not know what to talk to me or to anyone else.
في خميس، 18 مارس، 2021 في 7:04 م، كتب Hsiao-Wei Wang <
***@***.***>:

Issue #302, opened 2021-01-25
Add internationalization support


samajammin:
Related to #139, we'd like to add support for translations of the launchpad.

Here's my current thinking around sequential steps.

1. Configure the codebase to support translations (via `react-intl`)
2. Re-work app copy & design to avoid string fragments (which break translations)
3. Refactor all app copy into source language file (i.e. `src/intl/en.json`)
4. Upload source language file (`src/intl/en.json`) into Crowdin
5. Translate files in Crowdin
6. Add translated files (`src/intl/zh.json`, `src/intl/de.json`, etc.) to codebase
7. Add dropdown to site navigation / dedicated page that allows users to select from supported languages

I welcome input on this. I spiked on a quick prototype to add translation support & will open a PR for that shortly.

samajammin:
CC @EthereumCN - saw your comments in #139 & want to loop you in for feedback. I'm looking to help push this forward in the coming weeks. Are you available to collaborate on translation work? I'm thinking we take an approach similar to ethereum.org's by using Crowdin as the primary platform to manage translations.

Let me know what you think 😄 . If you're on board, we'd need some time (a week or 2?) to set up the content appropriately before uploading the content to Crowdin. Thanks!

I plan to manage the technical side of things to set up translation support in the codebase. @ryancreatescopy will lead the design & content effort to ensure the Launchpad content is 1) up to date prior to translating & 2) structured in a way that facilitates translations easily. @sorumfactory (who leads the Translation Program on ethereum.org) will be available to assist & answer questions related to the translation process. Together we plan to work with @CarlBeek to ensure we're taking all needs of the Launchpad into account.

samajammin:
FYI the long-running branch for this work can be found here: https://github.com/ethereum/eth2.0-deposit/pull/308

PR #303, opened 2021-01-25
[WIP] Add internationalization support

samajammin:
## Description
First step towards solving #302: add `react-intl` & configure codebase to support translations.

## TODO
- [ ] Configure the codebase to support translations
  - [x] Add [`react-intl`](https://www.npmjs.com/package/react-intl) & `src/intl/` file structure
  - [ ] Add [`react-helmet`](https://github.com/nfl/react-helmet) to set `lang` HTML attribute depending on route
- [ ] Re-work app copy & design to avoid string fragments (which break translations)
- [ ] Refactor all app copy into source language file (i.e. `src/intl/en.json`)


src/intl/de.json
```diff
@@ -0,0 +1,3 @@
+{
```
samajammin, line 1: Test file to ensure translation support works, this shouldn't actually be commited.

src/intl/fr.json
```diff
@@ -0,0 +1,3 @@
+{
```
samajammin, line 1: Test file to ensure translation support works, this shouldn't actually be committed.

samajammin:
Added react-helmet here: https://github.com/ethereum/eth2.0-deposit/pull/309

PR #307, opened 2021-02-01
Overview section

ryancreatescopy:
This PR addresses the main confirmations flow (with all the warnings and "I accept" buttons) when you follow the CTA to become a validator. 

Major updates include
- Giving the terminal its own step: You need to be able to use a terminal to run a validator
- Relabelling
- Adding a subtitle that explains the purpose of the section


ryancreatescopy:
Question:
- What is the main purpose of the `KeysSigning` step? I'm unsure whether this is something a user needs to know at this stage. It's less "something you should be aware of" and more just explaining an upcoming step. Unless I'm misunderstanding the point, maybe we should remove this step.

samajammin (approved):

PR #309, opened 2021-02-01
Add meta tags

samajammin:
Add `react-helmet` which supports translations for meta tags.


src/pages/FAQ/index.jsx
```diff
@@ -28,8 +29,12 @@ const SectionTitle = styled(Heading)`
 `;
 
 export const FAQ = () => {
+  const intl = useIntl();
   return (
-    <PageTemplate title="Validator FAQs">
+    <PageTemplate
+      title={intl.formatMessage({ id: 'page-faq-title' })}
```
samajammin, line 35: @ryancreatescopy FYI so here's how we'll primarily pass page title/description info - via the PageTemplate component.

PR #310, opened 2021-02-02
Language path adjustments

wackerow:
- Prevents `/en` from redirecting to `/en/en`
- If no two-letter language code, will default to `/en/`
- If two-letter code not supported, will redirect to `/en/lanagues` to select a supported language
- Cleaned up a couple console warnings (.map key/idx, and `<text>` -> `<p>`)


samajammin (approved):
👍

PR #313, opened 2021-02-02
Condensing steps

ryancreatescopy:
Made a few tweaks to combine steps in initial overview.


src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -116,35 +116,37 @@ export const pageContent = {
...
-      <Text size="large" className="my10">
-        We'll help you create keys for each of your validators. These keys will
-        be saved in key-stores and you will need to give these to your validator
-        software to begin validating. You will also receive a deposit file to
-        upload to this website with the public keys for your validator.
-      </Text>
+      <>
+        <Text size="medium" className="mt10">
+          To become a validator you'll need to know about managing keys. If
+          you're unfamiliar with keys and mnemomics, please don't proceed. Your
```
CarlBeek, line 124: I'm worried that users read this far, think "I know how mnemonics work" and stop reading further. I think we should be emphasising that this is slightly different to that in that the validator keys do now allow you to withdraw, only the mnemonic does that.

src/pages/Acknowledgements/AcknowledgementProgressTracker/index.tsx
```diff
@@ -22,9 +22,7 @@ const acknowledgementsWithCopy = {
   [AcknowledgementIdsEnum.deposit]: 'Deposit',
   [AcknowledgementIdsEnum.responsibilities]: 'Uptime',
   [AcknowledgementIdsEnum.slashing]: 'Bad behaviour',
-  [AcknowledgementIdsEnum.keyManagement]: 'Keys safety',
-  [AcknowledgementIdsEnum.signingKeys]: 'Keys signing',
-  [AcknowledgementIdsEnum.transferDelay]: 'Transfer delay',
+  [AcknowledgementIdsEnum.keyManagement]: 'Keys management',
```
samajammin, line 25:
```suggestion
  [AcknowledgementIdsEnum.keyManagement]: 'Key management',
```

samajammin: Would it be "key" here even with plural keys?

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -116,35 +116,37 @@ export const pageContent = {
...
+          will be saved in key-stores and you will need to give these to your
+          validator software to begin validating. You will also receive a
+          deposit file to upload to this website with the public keys for your
+          validator.
+        </Text>
+      </>
     ),
     acknowledgementText:
-      'I will safeguard my key-stores and give them to my validator software to start validating.',
+      'I understand that keys are my responsibility and that my mnemonic (seed) will be the ‼️ONLY WAY‼️ to withdraw my funds.',
```
samajammin, line 143:
```suggestion
      'I understand that my keys are my responsibility and that my mnemonic (seed) will be the ‼️ONLY WAY‼️ to withdraw my funds.',
```

samajammin: WE should come up with an alternative to `‼️ONLY WAY‼️` - I could see that creating confusion for translators. e.g. we could put this within bold HTML elements now that we know how.

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -157,38 +159,34 @@ export const pageContent = {
...
-    content: (
-      <Text size="large" className="my10">
-        With transfers disabled for now, you won't be able to voluntarily exit
-        and then restart later. This means you need to be in it for the long
-        haul.
-      </Text>
-    ),
     acknowledgementText:
-      'I understand that if I exit, I will not be able to rejoin until much later. This is a long term commitment.',
+      'I understand that I CANNOT TRANSFER my stake for a while, and I CANNOT WITHDRAW until the merge. I understand that if I exit, I will not be able to rejoin until much later. This is a long term commitment.',
```
samajammin, line 170: I'm not sure we should capitalize text like this. Not sure if it potentially creates confusion for translations. I'd lean towards assuming people will read this without the need for all caps.

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -157,38 +159,34 @@ export const pageContent = {
...
     title: 'Early adopter risks',
     content: (
-      <Text size="large" className="my10">
-        You'll be participating in the initial launch of a novel network. As
-        with any new piece of software, there is the potential for software
-        bugs. While unlikely, potential bugs may result in slashing.
-      </Text>
+      <>
+        <Text size="medium" className="my10">
+          You're joining a new network. As with any new piece of software, there
```
samajammin, line 177:
```suggestion
          You're joining a young network. As with any new piece of software, there
```

samajammin (commented):
I like this change for the most part. Left a few comments.

src/pages/Acknowledgements/pageContent.tsx
CarlBeek: Sounds reasonable. As long as it remains accentuated as this is different to how keys and accounts work on eth1.

PR #316, opened 2021-02-03
Translate languages page

samajammin:
Extract copy from languages page for translations

### DIFF
--- a/src/pages/Languages/index.tsx
+++ b/src/pages/Languages/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, useIntl } from 'react-intl';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/pages/Languages/index.tsx
```diff
@@ -109,8 +106,16 @@ export const Languages = () => {
         })}
       </LangContainer>
       <p>
-        If you'd like to see the launchpad in another language, or if you can
-        help translate, get in touch.
+        <FormattedMessage
```
samajammin, line 109: @ryancreatescopy @wackerow - we can use HTML elements & React components within translation phrases! 🤯 🎉

wackerow: Hm... so will the translator be able to see the whole context, and also be able to translate that variable (ie: `getInTouch`) so it stays a variable?
If so looks like this would work great, and could also be used to keep inline styling. =)

PR #319, opened 2021-02-03
Validating =\ mining

ryancreatescopy:
We've had a lot of confusion around whether staking makes you a miner in Eth2. This PR aims to combat this misconception.


samajammin (approved):
👍

PR #320, opened 2021-02-03
Deposit threshold and staking rewards

ryancreatescopy:
Removes references to "until threshold".
Adds simple context for the minimum threshold.

### DIFF
--- a/src/pages/Landing/NetworkStatus/index.tsx
+++ b/src/pages/Landing/NetworkStatus/index.tsx
@@ -3,28 +3,14 @@ import styled from 'styled-components';
 import ScrollAnimation from 'react-animate-on-scroll';
 import { Heading } from '../../../components/Heading';
 import { Text } from '../../../components/Text';
-import { ProgressBar } from './ProgressBar';
+import { Link } from '../../../components/Link';
+import { Button } from '../../../components/Button';
 import { numberWithCommas } from '../../../utils/numberWithCommas';
 import {
-  IS_MAINNET,
-  ETH_REQUIREMENT,
   ENABLE_RPC_FEATURES,
-  ETH2_NETWORK_NAME,
   PRICE_PER_VALIDATOR,
   TICKER_NAME,
 } from '../../../utils/envVars';
-import useMobileCheck from '../../../hooks/useMobileCheck';
-
-//
-// Helpers
-
-const calculatePercentage = (amountEth: number) => {
-  const percentage = (amountEth / +ETH_REQUIREMENT) * 100;
-  return percentage && Math.max(1, percentage);
-};
-
-const calculateLaunchThreshold = (amountEth: number) =>
-  Math.floor(+ETH_REQUIREMENT - amountEth);
 
 //
 // Styled Components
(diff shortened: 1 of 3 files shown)


ryancreatescopy:
@samajammin and @wackerow be curious for your thoughts on these copy changes. Wrote them earlier but wondering whether it's necessary at all now to mention the threshold.

src/pages/Landing/NetworkStatus/index.tsx
```diff
@@ -147,20 +136,21 @@ export const NetworkStatus: React.FC<{ amountEth?: number }> = ({
...
             <>
               <Text className="mt20">
-                The threshold to launch the eth2
-                {IS_MAINNET ? ` mainnet ` : ` ${ETH2_NETWORK_NAME} testnet `}
-                has been reached 🎉
+                {IS_MAINNET
+                  ? `The Beacon Chain `
+                  : ` ${ETH2_NETWORK_NAME} testnet `}{' '}
+                needs at least {numberWithCommas(ETH_REQUIREMENT)} {TICKER_NAME}{' '}
+                staked to be secure.
```
wackerow, line 153: @ryancreatescopy Overall I'd argue against this even being included anymore. If I understand correctly, the beacon chain will no longer achieve finality if the number of ETH staked goes below 524,288 (vs still operating insecurely as this kind of implies), but at this point it seems more of a technical nuance, and something people can read the docs to learn more about. 

Now, I would argue the total ETH staked is mostly all that matters, in relation to the reward they can expect when financially choosing whether or not to commit their ETH to staking. It is nice to see visually that we're not at the very beginning anymore, and I believe the chart does this very nicely, but showing where we are now, and leave a large gap on the chart for where we expect it to go.

src/pages/Landing/NetworkStatus/index.tsx
```diff
@@ -172,14 +162,9 @@ export const NetworkStatus: React.FC<{ amountEth?: number }> = ({
...
-                <BoldGreen className="mr10 ml10" fontSize={24}>
-                  {numberWithCommas(ETH_REQUIREMENT)} {TICKER_NAME},
-                </BoldGreen>
-                <BoldGray className="mr10" fontSize={24}>
-                  {validatorRequirement} validators,
-                </BoldGray>
-                to launch the
+                The Eth2 network needs at least{' '}
+                {numberWithCommas(ETH_REQUIREMENT)} {TICKER_NAME} (
+                {validatorRequirement} validators), to launch the
```
wackerow, line 167: Similarly, don't think this is needed anymore

ryancreatescopy:
Taking into account your comments @wackerow , I've removed the reference to the threshold at all. I think this could be confusing. 

I've replaced the progress bar with a stats section. @CarlBeek would be good to get your thoughts on this. It just shows the current APR, number of validators and total ETH staked at a glance. This could be expanded in the future to show other network metrics that we think potential stakers might find useful.

Removed a lot of components from the `NetworkStatus/Index` as they're no longer needed without the progressbar. I'd encourage us to check this PR thoroughly to make sure I haven't removed anything important.

PR #321, opened 2021-02-03
Adds mobile menu and shortcut to languages

ryancreatescopy:
Allows users to better access educational content when viewing on non-desktop devices.

Todo:
- [x] Adjust breakpoints: should mobile menu be used on tablet size devices?

### DIFF
--- a/src/components/AppBar.tsx
+++ b/src/components/AppBar.tsx
@@ -31,7 +31,7 @@ const EthLogo = styled.img`
 `;
 
 const NetworkText = styled(Text)`
-  padding: 5px 15px;
+  padding: 5px 8px;
   border-radius: 4px;
   border: 1px solid;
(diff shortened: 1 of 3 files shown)


wackerow:
@ryancreatescopy I feel like similar breakpoints as to the main repo would work fine (1024px), which with the current tabs may still be cutting it close.

samajammin:
Couple thoughts here
- Re: breakpoint, I'd support going larger (looks like they currently start hiding links at 1080px). Might as well transition to mobile nav here, so that all links remain accesible?
- I think it's worth keeping some text next to the logo to help explain the app & network, e.g. "Mainnet Eth2 Launchpad" & "Pyrmont Eth2 Launchpad". Something like that would remove the need to display the current network "button".

PR #329, opened 2021-02-04
Bypass acknowledgements while in dev mode

wackerow:
Changes default "acknowledgement" booleans to check for "development" mode, relieving workflow restrictions while in dev.
@ryancreatescopy @samajammin this should help


samajammin:
@wackerow I like this idea but when I try it, I still need to click through each acknowledgement. They're just all checked:
https://share.getcloudapp.com/eDu89ZZy

LMK if I'm missing something

wackerow:
Hm.. yeah, I am able to click directly to the last step on my end, but then you can't continue. Clicking second to last step can let you click through.. maybe not worth it, but I'll keep an eye out for a better way while working on stuff

samajammin:
Closing this out for now.

PR #331, opened 2021-02-04
pageContent updates and refactor with FormattedMessage

wackerow:
Affects "Overview" page with acknowledgments and disclaimers:
- Adds warning against storing keystores on multiple validators (currently the cause for 100% of slashings to date) with link to read more
- `b` -> `strong` for accessibility reasons
- Changed title of one section to "Key management" (from "Keys"), but this also brings up challenge with providing a translatable string to JS object, as it's currently typed as "string" from its interface. @samajammin Thoughts on this?
 
Also refactored strings to use `<FormattedMessage />`


samajammin (approved):
👍

PR #334, opened 2021-02-05
Checklist refactor using FormattedMessage

wackerow:
- Updated approximate Geth size
- Added recommendation to not wait on eth1 chain syncing
- Refactored Checklist to use `<FormattedMessage />` component and `formatMessage()`
- Provided descriptions whenever "values" attribute is used to assist translators. 

@samajammin Lemme know if this looks correct to prepare for crowdin

### DIFF
--- a/src/pages/Checklist/index.tsx
+++ b/src/pages/Checklist/index.tsx
@@ -198,7 +198,7 @@ export const Checklist = () => {
                     ancient: <em>ancient</em>,
                     chaindata: <em>chaindata</em>,
                   }}
-                  description="The {mainnet}, {ancient} and {chaindata} values are those same words but with emphasis styling"
+                  description="The {mainnet}, {ancient} and {chaindata} values are network and folder names that should not be translated"
                 />
               </Text>
             </li>
@@ -419,12 +419,18 @@ export const Checklist = () => {
             <li className="py5">
               <Text>
                 <FormattedMessage
-                  defaultMessage="{warning}: it's high risk to run your Eth2 validator in multiple places. It will lead to a slashable event and ejection from the network. {learnMore}"
+                  defaultMessage="{warning} It's high risk to run your Eth2 validator in multiple places. It will lead to a slashable event and ejection from the network. {learnMore}"
                   values={{
-                    warning: <strong>Warning: </strong>,
+                    warning: (
+                      <strong>
+                        {formatMessage({ defaultMessage: 'Warning!' })}
+                      </strong>
+                    ),
                     learnMore: (
                       <Link primary inline to="/faq#responsibilities">
-                        Learn more about slashing →
+                        {formatMessage({
+                          defaultMessage: 'Learn more about slashing →',
+                        })}
                       </Link>
                     ),
                   }}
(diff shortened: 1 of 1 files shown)


src/pages/Checklist/index.tsx
```diff
@@ -246,80 +286,137 @@ export const Checklist = () => {
...
               </Text>
             </li>
             <li className="py5">
               <Text>
-                <strong>Note:</strong> by default, VMs may disable NTP so you
-                may need to find a work-around for your environment.
+                <FormattedMessage
+                  defaultMessage="{note}: by default, VMs may disable NTP so you may need to find a work-around for your environment."
+                  values={{
+                    note: <strong>Note</strong>,
```
samajammin, line 370: This is tricky & something we may want to choose to just remove the bold on "Note".

If we want to keep the bold, we'd still want to translate "Note", so the value should be:
```
note: <strong>formatMessage({defaultMessage: "Note"})</strong>
```
But that would take "Note" out of context. Translators may thinks it's the noun vs. the verb.

samajammin: We could use a `description` to explain it's the verb but the lack of tense general obscurity might make it too confusing to translators for this to be worth it.

wackerow: Yeah, the description was how I was attempting to go about this. Removing the bold would likely be the best option. These aren't really warnings, and don't demand a high level of attention. 
And good call on using `formatMessage()` to translate within the `values` attribute... wasn't thinking, I can add another commit to fix any of these.

src/pages/Checklist/index.tsx
```diff
@@ -246,80 +280,127 @@ export const Checklist = () => {
...
-                <Code>Universal time</Code> are all correct.
+                <FormattedMessage
+                  defaultMessage="Check if {code1}, {code2}, and {code3} are all correct."
+                  values={{
+                    code1: <Code>Local time</Code>,
+                    code2: <Code>Time zone</Code>,
+                    code3: <Code>Universal time</Code>,
+                  }}
+                  description="{code} values are terminal outputs to look for to confirm system time"
+                />
```
wackerow, line 326: @samajammin honestly not sure if these should be translated or not... I would think a user's system settings would produce these in their native language, but not sure.

src/pages/Checklist/index.tsx
samajammin: I agree, in general I suspect the value of bolding words won't be worth the effort/potential confusion. I think this feature will be used best with variables that we do NOT want translated (such as code snippets / CLI commands).

src/pages/Checklist/index.tsx
samajammin: Good question. @CarlBeek do you know anyone who might know this? Does `timedatectl ` support multiple languages? A quick Google didn't yield any aswers.

samajammin: If in doubt I suggest we keep it in English.

wackerow: Okay... this is set up to not be translated at the moment, I will leave it pending any new information.

samajammin (approved):

src/pages/Checklist/index.tsx
samajammin: Merging this for now. @CarlBeek please let us know if you're aware of any `timedatectl` language support.

PR #335, opened 2021-02-05
Congratulations page refactor using FormattedMessage

wackerow:
Refactored "Congratulations" page with `react-intl`
Added descriptions where it felt helpful

### DIFF
--- a/src/pages/Congratulations/index.tsx
+++ b/src/pages/Congratulations/index.tsx
@@ -7,18 +7,17 @@ import { Text } from '../../components/Text';
 import { Alert } from '../../components/Alert';
 import { Link } from '../../components/Link';
 import { FormNext } from 'grommet-icons';
-import { colors } from '../../styles/styledComponentsTheme';
-import { ProgressBar } from './ProgresBar';
 import { queryContract } from '../../utils/queryContract';
-import { ProgressBarInfo } from './ProgressBarInfo';
 import { DepositKeyInterface, StoreState } from '../../store/reducers';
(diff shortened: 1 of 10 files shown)


src/pages/Congratulations/index.tsx
```diff
@@ -147,21 +149,31 @@ const _CongratulationsPage = ({
...
                     amountEth={depositKeys.length * PRICE_PER_VALIDATOR}
                     amountValidators={depositKeys.length}
                   />
                   <ProgressBarInfo
-                    title="Launch threshold:"
+                    title={formatMessage({
+                      defaultMessage: 'Launch threshold:',
+                      description:
+                        'Refers to threshold of ETH needed to launch the beacon chain',
+                    })} /* TODO: remove this reference? */
```
wackerow, line 176: @samajammin @ryancreatescopy Another instance of this that should be removed if you agree.

ryancreatescopy:
I worked on top of your fine work @wackerow. Again the progress bar seemed a little irrelevant with the threshold stuff. Some people may still like to see the percentage of their contributions against the total. 

I...
- introduced the cards we added to the homepage to provide clearer confirmation of the actions (e.g. amount staked and validators created)
- - Added a clearer CTA to visit the checklist
- added an info alert that explains the validator won't come online for a bit... encouraged users to go to the checklist and set up a testnet
- - need a way to dynamically display testnet name and the correct testnet URL in the alert.
- Revamped the banner at the bottom to once again advertise the checklist but also state that they can get support from EthStaker community
- - maybe we actually create two cards here to make the two actions more discreet.

PR #338, opened 2021-02-05
Connect wallet updates

ryancreatescopy:
This step focusses on the user choosing a wallet and getting feedback from the app whether they're on the correct network and they have the required balance for the number of validators they want to run.

Problems fixed:
- previous design relied on color to indicate whether the network was correct. This is inaccessible so added an icon to show status.
- it was confusing that the dot would display green (to demo that the network was correct) even if you couldn't proceed because of an insufficient balance. This is now split into two separate checks.
- Added error messages inline to wallet balance and network statuses
- disabled the continue step if balance is insufficient (I assume this was accidentally left out before – please correct me if this is incorrect)

Todo:
- [ ] format status message for translations
- [ ] format button copy for translations

### DIFF
--- a/src/pages/ConnectWallet/index.tsx
+++ b/src/pages/ConnectWallet/index.tsx
@@ -198,6 +198,7 @@ const _ConnectWalletPage = ({
         error.message === 'Invariant failed: chainId 0xNaN is not an integer')
     );
   }, [error]);
+  const intl = useIntl();
 
   // setup RPC event listener
   const attemptedMMConnection: boolean = useMetamaskEagerConnect();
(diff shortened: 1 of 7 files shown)


src/pages/ConnectWallet/index.tsx
```diff
@@ -190,12 +248,12 @@ const _ConnectWalletPage = ({
...
       (balance || balance === 0)
     ) {
-      setStatus(`${balance} ${TICKER_NAME} available`);
+      setStatus(`${balance} ${TICKER_NAME}`);
     } else if (walletConnected && error) {
       setStatus('Error');
     } else if (!networkAllowed) {
       setStatus(
-        `Please connect to ${
+        `Connect ${getWalletName(walletProvider)} to ${
```
ryancreatescopy, line 274: @samajammin @wackerow any idea how we set up this kinda string for translataions?

src/pages/ConnectWallet/index.tsx
```diff
@@ -234,63 +284,95 @@ const _ConnectWalletPage = ({
...
+                          {depositKeys} {validator}"
+                          values={{
+                            depositKeys: (
+                              <span>
+                                {depositKeys.length}
+                                {depositKeys.length > 1 ? 's' : ''}
+                              </span>
+                            ),
+                            eth: <span>{TICKER_NAME}</span>,
+                            validator: (
```
ryancreatescopy, line 370: and this one @samajammin @wackerow

PR #339, opened 2021-02-06
Eth2 client pages refactored using FormattedMessage

wackerow:
- Eth2 clients Teku, Nimbus, Prysm and Lighthouse pages refactored using `FormattedMessage` component
- Cleaned up some code strings


samajammin (approved):
Nice work @wackerow!

PR #340, opened 2021-02-06
Fix typo in FAQ

jakerockland:
"What happens I lose my signing key" -> "What happens *if* I lose my signing key"


samajammin (approved):
LGTM - @CarlBeek given that it's a PR into `dev` is this something you should review?

CarlBeek:
Indeed, thanks @jakerockland!

PR #344, opened 2021-02-08
Set up transactions page for translations

ryancreatescopy:
- Consolidated transactions "confirm all tx" and key files table into one section
- Adjusted some language
- Set up for translations

Note: it's kind of hard to test all the different outcomes, would be worth going through together @samajammin @wackerow  v carefully.

### DIFF
--- a/src/pages/Transactions/index.tsx
+++ b/src/pages/Transactions/index.tsx
@@ -57,6 +57,7 @@ const _TransactionsPage = ({
   dispatchTransactionStatusUpdate,
   dispatchWorkflowUpdate,
 }: Props): JSX.Element => {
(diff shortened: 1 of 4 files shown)


src/pages/Transactions/index.tsx
```diff
@@ -85,27 +86,27 @@ const _TransactionsPage = ({
 
   const createButtonText = (): string => {
     if (totalTxCount === 1) {
-      return 'Initiate the transaction.';
+      return 'Confirm deposit';
```
samajammin, line 89: These strings should be wrapped in `intl.formatMessage()`

wackerow: Will push a fix

PR #350, opened 2021-02-10
Key management overview step trimming

wackerow:
- Rearranged the "Key management" overview step copy to fit a little nicer and remove some redundancy. This prevents the "Continue" button from shifting downward too much during this flow.
- Also adds an instance of intl to acknowledgement index page

### DIFF
--- a/src/pages/Acknowledgements/pageContent.tsx
+++ b/src/pages/Acknowledgements/pageContent.tsx
@@ -155,33 +155,10 @@ export const pageContent = {
         </Text>
         <Text size="medium" className="mt20">
           <FormattedMessage
-            defaultMessage="Your validator keys are derived from a unique mnemonic (or seed). Your
-              mnemonic/seed will be the ONLY WAY to withdraw your funds. Above all, keep it safe! No
-              one can help you if you lose this phrase."
-          />
-        </Text>
-        <Text size="medium" className="mt20">
-          <FormattedMessage
-            defaultMessage="We'll help you create keys for each of your validators. These keys
-              will be saved in password-encrypted keystore files and you will need to give these
-              to your validator software to begin validating*. You will also receive a deposit file
-              to upload to this website with the public keys for your validator."
-          />
-        </Text>
-        <Text size="medium" className="mt20">
-          <FormattedMessage
-            defaultMessage="*Warning: Do not store keys on multiple/backup validators at once {readMore}"
-            values={{
-              readMore: (
-                <Link
-                  className="my10"
-                  primary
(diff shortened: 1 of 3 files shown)


src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -165,13 +170,7 @@ export const pageContent = {
...
         <Text size="medium" className="mt20">
           <FormattedMessage
-            defaultMessage="If you're unfamiliar with keys and mnemomics, please don't proceed. Your
-              keys will be your responsibility and no one can help you if you lose them."
-          />
-        </Text>
-        <Text size="medium" className="mt20">
-          <FormattedMessage
-            defaultMessage="*Warning: Do not store keys on multiple/backup validators at once. {readMore}"
+            defaultMessage="*Warning: Do not store keys on multiple/backup validators at once {readMore}"
```
ryancreatescopy, line 173: I wonder whether this warning would be better on the generate keys step. There's already a lot of information on this page.

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -147,12 +147,17 @@ export const pageContent = {
...
+            defaultMessage="To become a validator you'll need to know about managing keys and
+              protecting a mnemonic. If you are not yet familiar with keys and mnemomics, please
+              do not proceed."
+          />
         </Text>
         <Text size="medium" className="mt20">
           <FormattedMessage
-            defaultMessage="Your validator keys are derived from a uniquemnemonic (or seed). Your
-              seed will be the ONLY WAY to withdraw your funds. Above all, keep it safe!"
+            defaultMessage="Your validator keys are derived from a unique mnemonic (or seed). Your
```
ryancreatescopy, line 158: I wonder whether the paragraph below is needed in this section. 

```
"We'll help you create keys for each of your validators. These keys will be saved in password-encrypted keystore files and you will need to give these to your validator software to begin validating*. You will also receive a deposit file to upload to this website with the public keys for your validator."
```

Could it be something like: 

```
"We'll help you create keys for every validator you want to run. They keys are an important part of linking your ETH with your validator. They'll be generated via a mnemonic (seed) which you need to keep safe. Your mnemonic will be the only way to withdraw your ETH when the time comes and no one can help you recover your mnemonic if you lose it.
 ```
 
 This will condense the two paragraphs and remove redundant info (for now)

ryancreatescopy (changes requested):
See comments

ryancreatescopy (approved):

ryancreatescopy:
Approved as I understand the titles in `pageContent` are covered by `<FormattedMessage>` in an other PR.

samajammin (approved):

PR #352, opened 2021-02-10
Formatted message cleanup

ryancreatescopy:
PR to catch any content we haven't yet wrapped in `<FormattedMessage>`

### DIFF
--- a/src/pages/Acknowledgements/pageContent.tsx
+++ b/src/pages/Acknowledgements/pageContent.tsx
@@ -18,7 +18,7 @@ export interface PageContentInterface {
 
 export const pageContent = {
   [AcknowledgementIdsEnum.introSection]: {
-    title: 'Proof of stake',
+    title: <FormattedMessage defaultMessage="Proof of stake" />,
     content: (
       <>
         <Text size="medium" className="my10">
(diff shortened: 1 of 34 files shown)


wackerow:
@ryancreatescopy @samajammin I _think_ that's it for preparing strings for translation, but there were a lot so still a chance I missed one or two. Happy to review this with both/either of you together.

samajammin:
Great! I'll take a look tomorrow.

ryancreatescopy:
`GenerateKeys/index` and `Checklist` not touched as they're covered by other PRs

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -11,14 +11,14 @@ const BoldCaps = styled.strong`
...
+  title: JSX.Element;
   content: JSX.Element;
-  acknowledgementText?: any;
+  acknowledgementText?: JSX.Element;
 }
 
 export const pageContent = {
   [AcknowledgementIdsEnum.introSection]: {
-    title: 'Proof of stake',
+    title: <FormattedMessage defaultMessage="Proof of stake" />,
```
samajammin, line 21: @wackerow why not use the `formatMessage` function here so that the values could remain strings? Was it not possible to use the `useIntl()` hook?

samajammin (approved):
Nice work!

src/pages/Acknowledgements/pageContent.tsx
wackerow: @samajammin Exactly. Hooks need to be inside React components... `pageContent` is just an object.

PR #353, opened 2021-02-11
Simplify env vars

samajammin:
Small change but @CarlBeek would like to get your review to ensure this won't screw anything up e.g. on the Netlify side where we may set ENV variables.

Just trying to consolidate the places where we specify the testnet name & deduce from `IS_MAINNET` when possible so it's easier to update when the time comes.


samajammin:
Hey @CarlBeek checking in on this. Thanks!

src/utils/envVars.ts
```diff
@@ -7,7 +7,10 @@ export const ENABLE_RPC_FEATURES        = Boolean(INFURA_PROJECT_ID && INFURA_PR
 export const INFURA_URL                 = `https://${IS_MAINNET ? "mainnet" : "goerli"}.infura.io/v3/${INFURA_PROJECT_ID}`;
 
 // public
-export const ETH2_NETWORK_NAME          = process.env.REACT_APP_ETH2_NETWORK_NAME   || 'Mainnet';
+export const TESTNET_LAUNCHPAD_NAME      = 'Pyrmont'
+export const TESTNET_LAUNCHPAD_URL      = `https://${TESTNET_LAUNCHPAD_NAME.toLowerCase()}.launchpad.ethereum.org/`
+export const MAINNET_LAUNCHPAD_URL      = 'https://launchpad.ethereum.org/'
+export const ETH2_NETWORK_NAME          = IS_MAINNET ? TESTNET_LAUNCHPAD_NAME : 'Mainnet';
```
CarlBeek, line 13: I think this is backwards, shouldn't it be:
```suggestion
export const ETH2_NETWORK_NAME          = IS_MAINNET ? 'Mainnet': TESTNET_LAUNCHPAD_NAME ;
```

src/utils/envVars.ts
```diff
@@ -7,7 +7,10 @@ export const ENABLE_RPC_FEATURES        = Boolean(INFURA_PROJECT_ID && INFURA_PR
 export const INFURA_URL                 = `https://${IS_MAINNET ? "mainnet" : "goerli"}.infura.io/v3/${INFURA_PROJECT_ID}`;
 
 // public
-export const ETH2_NETWORK_NAME          = process.env.REACT_APP_ETH2_NETWORK_NAME   || 'Mainnet';
+export const TESTNET_LAUNCHPAD_NAME      = 'Pyrmont'
```
CarlBeek, line 10: I think this should still be configurable via a `process.env.` env-var as there are times when we have multiple concurrent testnets and ideally, the launchpad is not where this is configured. You shouldn't have to change the source code to change the testnet name.

CarlBeek (commented):
I do think there is some cleaning up to do here, but I am not convinced that this is the way to do it. IMO, you shouldn't have to go into the source code to configure a testnet name. (Especially as sometimes there are multiple concurrent testnets).

PR #356, opened 2021-02-12
Trims margin on workflow nav items

wackerow:
- Trims the horizontal margin on workflow nav items by half (20px -> 10px)
- Prevents wrapping of workflow nav item names above 1080px screen size

Bug ("Transaction" wrapping to new line):
image

Fix:
image

### DIFF
--- a/src/components/WorkflowPage/WorkflowProgressBar.tsx
+++ b/src/components/WorkflowPage/WorkflowProgressBar.tsx
@@ -16,7 +16,7 @@ const StepContainer = styled.div`
 `;
 
 const Step = styled.div`
-  margin: 0 20px;
+  margin: 0 10px;
   text-align: center;
   color: ${(p: {
     disabled: boolean;


ryancreatescopy:
One thing to bear in mind is this will almost definitely go onto two lines in a lot of languages. 

Question: how much value is there in knowing that there will be a summary in 3 steps time when you're on generate keys? Perhaps we could transition to a design that states like Step 3/7 and maybe it just displays the current step name and the next step name. There could be an option for the user to expand to see all steps.

wackerow:
@ryancreatescopy I kinda like the simplicity of that suggestion. Would instantly solve issues with some languages being larger than others

samajammin:
> One thing to bear in mind is this will almost definitely go onto two lines in a lot of languages.

@wackerow @ryancreatescopy do we want to move forward with this PR? Or close given this fact?

wackerow:
@ryancreatescopy Any opposition to just merging this to help at least the English headers? Or do you feel the margins are too tight now?

samajammin (approved):

samajammin:
Merging in... seems to at least be a win for English. @ryancreatescopy LMK what you think. We can revert if we don't want this.

PR #358, opened 2021-02-16
Restores active workflow path highlighting

wackerow:
- Intl upgrades altered the URL path being checked when deciding which active step the user is on
- This PR adds logic to check for the language locale and remove it from the path before checking it against the enumerated step names, and thus restores the conditional color formatting for the active item on the workflow nav bar. 

(@samajammin @ryancreatescopy)


samajammin (approved):
👍

PR #360, opened 2021-02-17
Links to ethereum.org maintain users locale

wackerow:
- Added logic to check for links going to `ethereum.org` pages (no sub-domains)
- Checks if link matches users current locale / language selection
- If it does, link passes unchanged. If it doesn't match or doesn't have a language path, the users current locale is injected to the link.

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -89,6 +89,14 @@ export const Link = (props: LinkProps) => {
   } = props;
   const { locale } = useIntl();
   const isExternal = to && to.includes('http');
+  // Check for ethereum.org root domain links
+  const ETHEREUM_DOT_ORG = 'ethereum.org';
+  const isEthereumDotOrg =
+    to &&
+    (to.includes(`http://${ETHEREUM_DOT_ORG}`) ||
+      to.includes(`https://${ETHEREUM_DOT_ORG}`) ||
+      to.includes(`http://www.${ETHEREUM_DOT_ORG}`) ||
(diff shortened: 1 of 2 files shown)


src/components/Link.tsx
```diff
@@ -89,6 +89,14 @@ export const Link = (props: LinkProps) => {
   } = props;
   const { locale } = useIntl();
   const isExternal = to && to.includes('http');
+  // Check for ethereum.org root domain links
+  const ETHEREUM_DOT_ORG = 'ethereum.org';
+  const isEthereumDotOrg =
```
samajammin, line 94: I think here we can just check if `to.includes('https://ethereum.org')`. Let's update any links that contain `http` or `www` - we shouldn't encourage that type of linking.

src/components/Link.tsx
```diff
@@ -106,10 +114,32 @@ export const Link = (props: LinkProps) => {
   }
 
   if (isExternal) {
+    let href = '';
+    if (isEthereumDotOrg) {
+      // Grab everything after "ethereum.org/"
+      const slug: string = to.substr(
+        to.indexOf(ETHEREUM_DOT_ORG) + ETHEREUM_DOT_ORG.length + 1
```
samajammin, line 121: If we just check for `const ETHEREUM_DOT_ORG = 'https://ethereum.org'`, then `to.indexOf(ETHEREUM_DOT_ORG)` will always be 0.

samajammin (changes requested):
Looks great! A couple small requests.

samajammin (approved):

PR #361, opened 2021-02-17
Pulls validator count / total eth staked from Beaconcha.in API

wackerow:
- Used a serverless lambda fuction (due to CORS issues) to call Infura Eth2 Beacon Chain endpoint to grab a more accurate validator count
- Have not been able to test yet (blocked by netlify access)

Does require a project ID/secret from Infura (basic auth)

**EDIT: Revamped to use **Beaconcha.in** API endpoint... see comment below**


samajammin:
This is dependent on https://github.com/ethereum/devops/issues/629 to get a Netlify function up & running

wackerow:
- Switched this to use Beaconcha.in API endpoint to get around CORS issue. Working locally for me, both mainnet and testnet.
- As a result, no more need for basic auth, nor serverless function

src/pages/Landing/NetworkStatus/index.tsx
```diff
@@ -114,7 +106,11 @@ export const NetworkStatus: React.FC<{ amountEth?: number }> = ({
               </Heading>
               <Text size="x-large" className="mt20">
                 <BoldGreen className="mr10" fontSize={24}>
-                  {validatorAmount}
+                  {totalValidators === 0 ? (
```
samajammin, line 109: This logic (`totalValidators === 0`) to determine if the query is loading might be problematic because we return 0 if there's an error with the API.

samajammin (changes requested):
Nice work!

Let's update our logic slightly to account for API errors, e.g. by including a loading/error state to react to vs. relying on validator count to equal 0.

src/pages/Landing/NetworkStatus/index.tsx
wackerow: Will take a look at this now

src/utils/envVars.ts
```diff
@@ -1,10 +1,12 @@
 export const IS_MAINNET                 = Boolean(process.env.REACT_APP_IS_MAINNET !== 'false');  // If REACT_APP_IS_MAINNET is unset, set it to true by default
+export const TESTNET_LAUNCHPAD_NAME      = 'Pyrmont'
 
 // private vars (or derived from)
 export const PORTIS_DAPP_ID             = process.env.REACT_APP_PORTIS_DAPP_ID     || '';
 export const INFURA_PROJECT_ID          = process.env.REACT_APP_INFURA_PROJECT_ID  || '';
 export const ENABLE_RPC_FEATURES        = Boolean(INFURA_PROJECT_ID && INFURA_PROJECT_ID !== '');
 export const INFURA_URL                 = `https://${IS_MAINNET ? "mainnet" : "goerli"}.infura.io/v3/${INFURA_PROJECT_ID}`;
+export const BEACONCHAIN_ENDPOINT       = `https://${IS_MAINNET ? `` : `${TESTNET_LAUNCHPAD_NAME.toLowerCase()}.`}beaconcha.in/api/v1/epoch/latest`;
```
samajammin, line 9: We already have a `BEACONCHAIN_URL` variable in this file. Let's just use that.

e.g. so in queryBeaconchain.ts:
```
axios.get(`${BEACONCHAIN_URL}/api/v1/epoch/latest)
```

wackerow: Good call, fixing now.

samajammin (approved):
Nice work 👍

PR #362, opened 2021-02-18
Compile translations

samajammin:
Add `compile` script to properly compile & import translation JSON files. 

This should resolve the host of console errors we've been seeing.

See https://github.com/formatjs/formatjs/issues/2589 for more context


samajammin:
FYI @wackerow - something to note, we'll want to compile these JSON files & ultimately import messages from `/src/intl/compiled/${lang}.json`.

PR #365, opened 2021-02-22
Fix external links

samajammin:
Fix bug from #360 where all non-ethereum.org external links were set to empty string.


wackerow (approved):
@samajammin Good catch, my bad. Neglected that this was within the `isExternal` check... patch looks good 👍

samajammin:
All good! I didn't catch it in my review either 😅

PR #366, opened 2021-02-22
Post meeting fixes

samajammin:


samajammin:
@wackerow note: there's more uses of `Mainnet` in some env template files (e.g. `.env.mainnet`).

samajammin:
@wackerow looks like this is ready for review now? LMK if there's any remaining todos surfaced from our call this morning.

wackerow:
The only thing I can think of is the WalletConnect component, which doesn't update automatically if you go from insufficient funds to enough funds. User still has the option to go back then forward again in the workflow to get it to kick in. Everything else should be ready for review.

samajammin:
@wackerow nice! Will take a look tomorrow. I just pushed a WIP implementation to address that WalletConnect issue: #369

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -153,9 +153,12 @@ export const pageContent = {
               do not proceed."
           />
         </Text>
+        <Text size="medium" className="mt20">
+          <FormattedMessage defaultMessage="We'll help you create signing keys for every validator you want to run and your withdrawal keys. Signing and withdrawal keys are separate by design. Your signing key must be held online because it needs to be available at all times. Your withdrawal keys don't need to be online so it's better to keep them separate for security." />
```
samajammin, line 157: > We'll help you create signing keys for every validator you want to run and your withdrawal keys.

This feels awkward. Perhaps instead:

> We'll help you create signing keys and withdrawal keys for every validator you want to run.

samajammin: Also, should this be singular @wackerow? i.e. 

> We'll help you create a signing key and a withdrawal key for every validator you want to run.

src/pages/Congratulations/index.tsx
```diff
@@ -1,13 +1,21 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
```
samajammin, line 1:
```suggestion
import React, { useEffect, useState } from 'react';
```
We still use these hooks

src/pages/Acknowledgements/pageContent.tsx
samajammin, line 157:
```suggestion
          <FormattedMessage defaultMessage="We'll help you create a signing key and a withdrawal key for every validator you want to run. Signing and withdrawal keys are separate by design. Your signing key must be held online because it needs to be available at all times. Any withdrawal keys don't need to be online so it's better to keep them separate for security." />
```

src/pages/Acknowledgements/pageContent.tsx
```diff
@@ -153,9 +153,12 @@ export const pageContent = {
               do not proceed."
           />
         </Text>
+        <Text size="medium" className="mt20">
+          <FormattedMessage defaultMessage="We'll help you create a signing key and a withdrawal key for every validator you want to run. Signing and withdrawal keys are separate by design. Your signing key must be held online because it needs to be available at all times. Your withdrawal key doesn't need to be online so it's better to keep them separate for security." />
```
samajammin, line 157:
```suggestion
          <FormattedMessage defaultMessage="We'll help you create a signing key and a withdrawal key for every validator you want to run. Signing and withdrawal keys are separate by design. Your signing key must be held online because it needs to be available at all times. Your withdrawal key doesn't need to be online so it's better to keep it separate for security." />
```

src/pages/Congratulations/index.tsx
```diff
@@ -146,29 +194,59 @@ interface StateProps {
...
+
+  const allTxConfirmed = _every(
+    depositKeys.map(
+      file => file.transactionStatus === TransactionStatus.SUCCEEDED
+    )
+  );
+
+  const actualTxConfirmed = totalTxCount - remainingTxCount;
+
+  const currentAPR = calculateEth2Rewards({ totalAtStake: actualTxConfirmed });
```
samajammin, line 236: @wackerow `totalAtStake` be the total ETH balance across all validators (like on the homepage, which we query from Beaconcha.in)?

wackerow: Yep, good catch 👍... that explains the weird APR we were getting at the end.

PR #369, opened 2021-02-24
Poll account balance

samajammin:
Subscribe to new blocks & update account balance if it changed


samajammin:
@wackerow @ryancreatescopy I'd like to abstract this functionality of fetching & setting balance into a function (since it's largely duplicate code from the other `useEffect` but the functionality should work.

PR #308, opened 2021-02-01
Translate the Launchpad

samajammin:
Add translation support!

While going through the process of configuring the codebase to support translations, we ended up making some significant design & content changes to the app 😅 

Fixes https://github.com/ethereum/eth2.0-deposit/issues/302

PR #371, opened 2021-02-25
Final adjustments (translations refactor)

wackerow:
Implements changes based on @hwwhww 's review. Hoping for this to be the last PR to collect any loose ends on `translations` branch.
- [x] Removes json-rpc info for eth1 clients while on testnet
- [x] removed GoETH usage on phishing page
- [x] ETHER_TO_GWEI utilized instead of 10**9
- [x] "Confirm deposits" -> "Send deposits" 
- [x] envVar improvements
- [x] Removed `Congratulations/ProgressBar` and `Congratulations/ProgressBarInfo` components; no longer being used after translation refactor

@hwwhww Held on swapping out TESTNET_LAUNCHPAD_NAME for ETH2_NETWORK_NAME (added comments, but basically because in those cases these will be equivilent, and I felt using TESTNET_LAUNCHPAD_NAME helped emphasize that these cases will never say "mainnet". Again, front end shouldn't change either way, and happy to add this in if you'd prefer. 
Also, you had made the comment about "total eth staked" which I replied to, please let me know if that makes sense and answers your question or not. 

cc @samajammin @ryancreatescopy


src/utils/envVars.ts
```diff
@@ -47,7 +47,10 @@ if(process.env.REACT_APP_PRICE_PER_VALIDATOR && Number.isNaN(Number(process.env.
 }
 export const PRICE_PER_VALIDATOR        = process.env.REACT_APP_PRICE_PER_VALIDATOR || 32;
 
-export const EJECTION_PRICE = 16;
+if(process.env.REACT_APP_EJECTION_PRICE && Number.isNaN(Number(process.env.REACT_APP_EJECTION_PRICE))) {
```
samajammin, line 50: Should we add this same check for `PRICE_PER_VALIDATOR`?

src/pages/Clients/Eth1/Besu.tsx
```diff
@@ -73,24 +73,28 @@ export const BesuDetails = () => (
...
-        />
-      </Text>
-      <Link
-        primary
-        to="https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-enabled"
-        className="mt20"
-      >
-        <FormattedMessage defaultMessage="rpc-http-enabled documentation" />
-      </Link>
+      {IS_MAINNET && (
```
samajammin, line 76: Are you sure this should only be shown if on mainnet?

I think the point was that it appears like it's under the "Testing on Goerli" header when that section is present:
Screen Shot 2021-02-25 at 6 09 03 PM

Perhaps @hwwhww can clarify their comment on https://github.com/ethereum/eth2.0-deposit/pull/308#discussion_r582951631

src/pages/Clients/Eth1/Geth.tsx
```diff
@@ -65,23 +65,27 @@ export const GethDetails = () => (
...
-          description="{http} shows '--http' terminal command"
-        />
-      </Text>
-      <Link
-        primary
-        to="https://geth.ethereum.org/docs/interface/command-line-options"
-      >
-        <FormattedMessage defaultMessage="--http documentation" />
-      </Link>
+      {IS_MAINNET && (
```
samajammin, line 68: Same question as above.

src/pages/Clients/Eth1/Nethermind.tsx
```diff
@@ -86,24 +86,28 @@ export const NethermindDetails = () => (
...
-        />
-      </Text>
-      <Link
-        primary
-        className="mt20"
-        to="https://docs.nethermind.io/nethermind/ethereum-client/json-rpc"
-      >
-        <FormattedMessage defaultMessage="--JsonRpc.Enabled documentation" />
-      </Link>
+      {IS_MAINNET && (
```
samajammin, line 89: Same question as above.

src/pages/FAQ/index.jsx
```diff
@@ -58,7 +58,7 @@ export const FAQ = () => {
             </Text>
             <Text className="mt10">
               <FormattedMessage
-                defaultMessage="Or in plain english, a human running a computer process. This
+                defaultMessage="Or in other words, a human running a computer process. This
```
samajammin, line 61: Nice catch 😄

samajammin (changes requested):
Nice! Just a few small comments/questions.

src/pages/Clients/Eth1/Besu.tsx
hwwhww: Ah sorry, I didn't clarify enough.
I meant the json-rpc setting could be set for both mainnet or goerli.

So IMO the headers could be either (1):

### Configurations
Use `--config goerli` ....
Use `--JsonRPC.Enabled true`...

---
Or (2):

### Testing on Goerli
Use `--config goerli` ....
### JSON RPC endpoint
Use `--JsonRPC.Enabled true`...

src/utils/envVars.ts
wackerow: Should be right above it already (L45-47), or am I missing something

src/utils/envVars.ts
```diff
@@ -1,5 +1,5 @@
 export const IS_MAINNET                 = Boolean(process.env.REACT_APP_IS_MAINNET !== 'false');  // If REACT_APP_IS_MAINNET is unset, set it to true by default
-export const TESTNET_LAUNCHPAD_NAME      = 'Pyrmont'
+export const TESTNET_LAUNCHPAD_NAME     = process.env.REACT_APP_TESTNET_LAUNCHPAD_NAME || 'Pyrmont';
```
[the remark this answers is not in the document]
wackerow: Yeah, that was the only reason, to be able to navigate between them. This functionality didn't exist before (ie, no links to Pyrmont from mainnet)

More that I look at this, I do not see a reason to define `REACT_APP_ETH2_NETWORK_NAME` in the `.env` file at all. We already have a `REACT_APP_IS_MAINNET` boolean, which if true (or if `REACT_APP_TESTNET_LAUNCHPAD_NAME` isn't available) can force `ETH2_NETWORK_NAME` to "mainnet", like:
```
export const ETH2_NETWORK_NAME = (IS_MAINNET || !process.env.REACT_APP_TESTNET_LAUNCHPAD_NAME) ? 'mainnet' : process.env.REACT_APP_TESTNET_LAUNCHPAD_NAME;
```

We already build in logic looking for the word "mainnet":
```
process.env.REACT_APP_ETH2_NETWORK_NAME.toLowerCase() === "mainnet"
```
...then force this variable to "mainnet", the above would remove the need for `REACT_APP_ETH2_NETWORK_NAME` in `.env`

I have these changes working locally already. Thoughts @hwwhww @samajammin ?

src/pages/Clients/Eth1/Besu.tsx
wackerow: Gotcha! Should've clarified, patching this up now

src/utils/envVars.ts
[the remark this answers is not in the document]
samajammin: @wackerow I hear ya, I opened a similar PR. See: https://github.com/ethereum/eth2.0-deposit/pull/353

Carl's main reason is to be able to support multiple testnets.

src/utils/envVars.ts
samajammin: Aha, whoops my bad. The Github PR didn't show those lines 😅

samajammin (changes requested):
@wackerow one more piece - run `yarn extract` && `yarn compile` & commit the generated output.

We need to figure out the best way to automate that 🤔  perhaps some sort of pre-commit hook.

src/utils/envVars.ts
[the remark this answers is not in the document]
wackerow: @samajammin Pushed this in 59fd7dd per our offline conversation, but just to document...

The method above should not prevent using a different testnet down the line. If Pyrmont is joined by a new testnet, the new testnet information can still be placed in the `.env` file, and it will replace any instance of "Pyrmont" on the page. This approach allows us to keep the new feature of being able to switch between networks, and removes redundant entries for the name of "mainnet", when "REACT_APP_IS_MAINNET" is enough.

wackerow:
@samajammin Added a pre-commit hook by injecting `"yarn extract", "yarn compile", "git add src/intl/*",` at the beginning of the `"lint-staged"` pre-commit hook array. Note, running this without the `git add src/intl/*` command left the language files unstaged and uncommitted.

Specifically, this should only run if commit includes `"*.{js,jsx,ts,tsx}"` files. You may notice 8d11d49 commits this change, but the language JSON was not extracted, because this commit did not include any changes to js/jsx/ts/tsx files. The following commit 59fd7dd affected `envVars.ts` so the extraction was triggered, and the latest language JSONs are included in this commit.

src/utils/envVars.ts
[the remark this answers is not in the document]
hwwhww: Note that we have to ask DevOps team (Jamie) to add the new env vars on Mainnet and Pyrmont servers **before** we merge the code into `master` branch.
/cc @CarlBeek

hwwhww (approved):
Looks good for me :)

samajammin (approved):
Thanks @wackerow for making these updates (& @hwwhww for the original review)!

PR #376, opened 2021-03-09
Use _blank link during the acknowledgements step

hwwhww:
### Issue

During the acknowledgements step, if the user clicks the internal link and then goes back, they will return to Step 1. For better UX, the links should be opened on new tabs.

### How did I fix it

Add `shouldOpenNewTab={true}` to these links.


samajammin (approved):
Good call!

PR #377, opened 2021-03-09
Fix links to beaconcha.in validator page

hwwhww:
### Issue

The URLs to beaconcha.in validator page was wrong.

### How did I fix it

Add resource name `validator/`.


samajammin (approved):
Thanks for catching!

Looks to be a bug from #308 - we updated `BEACONCHAIN_URL` to base URL in order to use more broadly but apparently we missed a couple items 😞

PR #380, opened 2021-03-15
Translation audit fixes

samajammin:
Various fixes identified from QA on https://pyrmont.launchpad.ethereum.org/


CarlBeek (approved):
Some nice fixes in here, good job, @wackerow, @samajammin & @hwwhww. 👍 

I am happy to merge this but would like to draw attention to something that we will probably end up having to deal with in the future. In https://github.com/ethereum/eth2.0-deposit/pull/380/commits/2008e1eb34ef86f7c5e03232325e9596019e8e8a, @wackerow added a case for single vs multiple deposits which is cool, but we're going to encounter similar, yet more complicated versions for some of the other languages. eg:

Some examples:
* In French zero is singular (zero deposit vs zero deposits)
* in some languages there are different words for singular, dual, and plural (as opposed to just singular and plural)
* Polish has different words for if the number of items ends in 2, 3 or 4, or some other number.
* Romanian has a different word if you have more than 20 of something

hwwhww (approved):
LGTM 👍

PR #384, opened 2021-03-16
Update top up link margins

samajammin:
Minor style tweak - unifies margin of "Top Up" nav link with the rest.

Context: we tightened the margins (https://github.com/ethereum/eth2.0-deposit/commit/1d0b6d6673e29a0279c04fdc8a36e18ba7999d92) but must have missed updating the Top Up link when we added that (https://github.com/ethereum/eth2.0-deposit/pull/291)


hwwhww (approved):
Nice catch!  🦅👁️  

The preview site:

image

PR #385, opened 2021-03-18
Fix Checklist dates

dxu429:
The string values passed to `FormattedDate` aren't parsed into valid Dates and is showing "Invalid Date" on the Checklist page.


samajammin (approved):
Thanks for catching this @dxu429!

Chrome & Brave seem to work fine but I can reproduce this issue on both Safari & Firefox browsers. This change solves them 🎉

PR #393, opened 2021-03-29
CLI flag translation prevention

wackerow:
At this moment, the Launchpad site is ready for translations, but the translated `--chain` CLI command shown to users is currently not a live feature in the actual deposit CLI tool. 

This adds a boolean conditional to skip translating this for now, and to just leave `--chain` in English, which the deposit CLI will accept without issues.

### DIFF
--- a/src/pages/GenerateKeys/Option2.tsx
+++ b/src/pages/GenerateKeys/Option2.tsx
@@ -144,6 +144,8 @@ export const Option2 = ({
   };
 
   const renderDepositKeyCommand = () => {
+    const translateFlags = false;
+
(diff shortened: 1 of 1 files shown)


samajammin (approved):

samajammin:
Good call 👍

hwwhww:
Sorry for the late review!

As we discussed in https://github.com/ethereum/eth2.0-deposit-cli/pull/182, it may be too confusing to use translated flags while the commands are not translated. IMHO we should keep `translateFlags = False`

CarlBeek:
I agree with not translating the flags.

PR #396, opened 2021-03-30
Add "withdrawal credentials" section to FAQ

hwwhww:
In that next deposit-cli release, https://github.com/ethereum/eth2.0-deposit-cli/pull/195 will introduce a new withdrawal credentials format - Eth1 address withdrawal. This PR adds some description about what are withdrawal credentials and this new feature.

Note that do not merge this PR before the next deposit-cli release.

### DIFF
--- a/src/intl/compiled/en.json
+++ b/src/intl/compiled/en.json
@@ -485,16 +485,6 @@
       "value": "Install deposit-cli tool"
     }
   ],
-  "3Bo3k6": [
-    {
-      "type": 1,
-      "value": "withdrawalCredentials"
-    },
-    {
-      "type": 0,
-      "value": " field is the 32-byte value in the deposit, for verifying the destination of valid withdrawals. Currently, there are two types of withdrawals: BLS withdrawal and Eth1 address withdrawal."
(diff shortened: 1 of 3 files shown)


src/intl/compiled/en.json
```diff
@@ -479,6 +485,16 @@
...
     }
   ],
+  "3Bo3k6": [
+    {
+      "type": 1,
+      "value": "withdrawalCredentials"
+    },
+    {
+      "type": 0,
+      "value": " field is the 32-byte value in the deposit, for verifying the destination of valid withdrawals. Currently, there are two types of withdrawals: BLS withdrawal and Eth1 address withdrawal."
```
CarlBeek, line 495: Grammatically something is off to me here:

"Withdrawal Credentials field is the 32-byte value in the deposit, for verifying the destination of valid withdrawals."

What about something like:
* "Withdrawal Credentials is a 32-byte field in the deposit"
* "The Withdrawal Credentials field is the 32-byte value in the deposit"

CarlBeek (approved):

PR #397, opened 2021-03-31
Remove incorrect "not"

rootulp:
During this step, you do derive the keys needed to withdraw later.
I don't think the word "not" belongs here.


hwwhww:
Hi!
I think it means that deposit-cli only generates the "signing key keystore", NOT "withdrawal key keystore".

By default, the withdrawal key can be generated with the same mnemonic in [EIP-2334 format](https://eips.ethereum.org/EIPS/eip-2334#eth2-specific-parameters).

PR #407, opened 2021-05-22
Fix typo

rootulp:


wackerow (approved):
@rootulp Thanks! 👍🏻

wackerow:
@samajammin This looks simple enough, but I thought about it more after approving... will this cause translations to break? If I recall, the english string is essentially hashed, and then it looks up the translation by that hash. In this case, that hash will now be different and it won't know which translated string to inject. If I'm correct we may need to update the reference to this string in the compiled intl files with the new hash.

samajammin (approved):
Thanks for the fix @rootulp!

Also nice catch to update the string hashes @wackerow 👍

rootulp:
No problemo. Thanks for all your hard work maintaining this project @samajammin @wackerow

PR #415, opened 2021-07-06
Sorts languages alphabetically

wackerow:


src/pages/Languages/index.tsx
```diff
@@ -142,7 +142,13 @@ export const Languages = () => {
       title: formatMessage({ defaultMessage: 'Polish' }),
       language: 'Polskie',
     },
-  ].filter(lang => supportedLanguages.includes(lang.code));
+  ]
+    .filter(lang => supportedLanguages.includes(lang.code))
+    .sort((a, b) => {
```
samajammin, line 147: I think best approach here may be `localCompare` as it handles several edge cases with string comparison:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

samajammin (approved):

src/pages/Languages/index.tsx
wackerow: Good call, thanks for fixing that and for the reference

PR #428, opened 2021-09-28
Renaming eth1/eth2 -> PoW/PoS or Execution/Consensus

hwwhww:
Changes:
- Launchpad wording as https://notes.ethereum.org/@timbeiko/great-renaming suggestions
- Since PoW/PoS chains haven't merged yet, this PR applies PoW/PoS more than Execution/Consensus.
- This PR only changes the wording. Not including (i) renaming `deposit-cli` and (ii) renaming parameters and file paths.
- This PR doesn't include the translations update. For example, the Chinese version would look like this 💦:
    image


src/pages/FAQ/index.jsx
```diff
@@ -628,7 +628,7 @@ export const FAQ = () => {
               <li>
                 <Text className="mt10">
                   <FormattedMessage
-                    defaultMessage="Eth1 address withdrawal: If you want to withdraw to Eth1 chain after the merge, you can set {eth1AddressWithdraw} when running deposit-cli. {boldWarning}"
+                    defaultMessage="Eth1 address withdrawal: If you want to withdraw to PoW chain (Eth1) after the merge, you can set {eth1AddressWithdraw} when running deposit-cli. {boldWarning}"
```
hwwhww, line 631: Note: `--eth1_withdrawal_address` is a spec parameter name. So, here I try to let users understand Eth1 address means PoW chain address.

wackerow:
If I may, I'm going to put this in draft mode for the moment to prevent it from being merged just yet...

With how the site is set up for translations, any changes to the english copy will change the hash of the lookup key associated with it. The means when a translated page is loaded after the english has been changed, the new lookup key will not be found in the languages `json` file. (See #429 for a recent example of this)

I am happy to update this with the necessary fixes, but I want to make sure we're all good with the changes to the english copy first before going ahead and adding these fixes. If we are, just let me know and I'll push a fix for the translated pages.

wackerow:
Hey @hwwhww, sorry for the delays here. So I do apologize that after finishing up work on the eth2 rebranding changes on the ethereum-org-website repo, @corwintines and I kept the momentum and went through this repo as well and put up #441, while this PR was still waiting here. 

---

The main differences I noted between our approaches was we used the terms "execution" and "consensus" much more, and didn't really utilize "PoW" at all...

My thoughts on this are that "execution layer" and "consensus layer" can properly delineate these two chains both now, and after the merge. Felt this would prevent the need to rework this specific content again once the merge occurs. Also feel these terms are more appropriate to sum up the chains they're referring to, instead of naming each chain by how block proposals are decided right now, in the case of using PoW or PoS.

Also think this will help people in general conceptualize the merge, since we're not merging PoW with PoS, but we are merging the execution layer with a new consensus layer. 

---

To your questions above, I'd suggest instead of a new branch from dev, we just use this `rebrand` branch for that purpose. 

For this specific change, why don't I change my PR from `dev` to `rebrand` (this branch), and if you wouldn't mind reviewing that and letting me know your thoughts, then we can pick and choose how to blend the two. When we're clear on copy changes I'll update the key hashes in the intl json files, fix any merge conflicts with `dev` and then we can bring it in.

samajammin (approved):
Awesome job @hwwhww @wackerow!

I made a few minor changes, mostly fixing links. I think this is ready to go!

@hwwhww I'll wait on you to give a final review & merge if you believe it's ready.

hwwhww (commented):
lgtm! thanks @wackerow @samajammin  <3

PR #444, opened 2022-01-27
Remove status

samajammin:
Reading "This Launchpad is in active development for testnet" gave me the impression that we're saying it's not ready for use on Mainnet. I don't think this is the message we want to send, given this app is used on Mainnet.

Am I misunderstanding this? Otherwise I think we should remove this statement.


hwwhww (approved):
good catch!

Issue #427, opened 2021-09-26
Encourage new validators to use a minority client


hesom:
Right now [approximately 65% of validators](https://twitter.com/sproulM_/status/1440512518242197516) run Prysm. In my opinion there should be a disclaimer about the importance of client diversity (not only for the network health but also because of the increased financial risk due to inactivity penalties if all Prysm validators went down). Maybe there could even be a warning if the user chooses Prysm

paulmillr:
I think it's very important to emphasize this. Totally agree

PR #457, opened 2022-02-11
Add client diversity warning if user selects Prysm

hesom:
Like I mentioned in #427, I think it is important to inform users of the risks of running a (super-) majority consensus client.
I added a simple disclaimer if the user chooses Prysm and a link to the Ethereum docs about client diversity.
The warning is hardcoded in the Prysm markup which means it has to be manually moved if another client gets majority in the future, but I don't see this happening anytime soon.

I don't know if this is even something you want on the website in this form or another but I'm open for suggestions.

disclaimer

### DIFF
--- a/src/pages/Clients/Consensus/Prysm.tsx
+++ b/src/pages/Clients/Consensus/Prysm.tsx
@@ -23,19 +23,26 @@ const ClientDiversityWarning = styled(Text as any)`
 export const PrysmDetails = ({ shortened }: { shortened?: boolean }) => (
   <>
     <ClientDiversityWarning>
-      <FormattedMessage
-        defaultMessage="Currently the majority of validators run Prysm as their consensus client.
-      Client diversity is extremely important for the network health of Ethereum:
-      A bug in a client with a share of over 33% can cause Ethereum to go offline. If the client has
-      a super-majority (>66%), a bug could cause the chain to incorrectly split, potentially leading to
-      slashing."
-      />
-      <Link
-        to="https://ethereum.org/en/developers/docs/nodes-and-clients/client-diversity/"
-        primary
-      >
-        <FormattedMessage defaultMessage="More on client diversity" />
-      </Link>
+      <p>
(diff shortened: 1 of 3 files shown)


src/pages/Clients/Consensus/Prysm.tsx
```diff
@@ -11,9 +12,31 @@ import { Link } from '../../../components/Link';
...
+      A bug in a client with a share of over 33% can cause Ethereum to go offline. If the client has
+      a super-majority (>66%), a bug could cause the chain to incorrectly split, potentially leading to
+      slashing."
+      />
+      <Link
+        to="https://ethereum.org/en/developers/docs/nodes-and-clients/client-diversity/"
+        primary
+      >
+        <FormattedMessage defaultMessage="More on client diversity" />
+      </Link>
```
wackerow, line 38: I like this ❤️... two small notes, "supermajority" doesn't need to be hyphenated, and I just suggested adding one last callout encouraging people to chose another client. Feel free to reword or ignore, just my thoughts. 

```suggestion
      <p>
        <FormattedMessage
          defaultMessage="Currently the majority of validators run Prysm as their consensus client.
            Client diversity is extremely important for the network health of Ethereum:
            A bug in a client with a share of over 33% can cause Ethereum to go offline. If the client has
            a supermajority (>66%), a bug could cause the chain to incorrectly split, potentially leading to
            slashing."
        />
      </p>
      <p>
        <FormattedMessage
          defaultMessage="If at all possible, consider running another client at this time to help protect yourself and the network."
        />
      </p>
      <p>
        <Link
          to="https://ethereum.org/en/developers/docs/nodes-and-clients/client-diversity/"
          primary
        >
          <FormattedMessage defaultMessage="More on client diversity" />
        </Link>
      </p>
```

wackerow (changes requested):
🙏🏻 Thanks for this! Let me know your thoughts on that comment

PR #463, opened 2022-02-28
Bulgarian, Hungarian, Slovenian - Import from Crowdin

wackerow:
## Description
Imports latest content from Crowdin for:
- Bulgarian (bg)
- Hungarian (hu)
- Slovenian (sl)

Noting the latest "eth2 rebranding" updates do not appear to be fully translated just yet. Open for feedback if we should wait or not.


samajammin:
> Noting the latest "eth2 rebranding" updates do not appear to be fully translated just yet.

Then why are we uploading these? Did Acolad review them prior to those updates?

wackerow:
@samajammin Not sure, I had grabbed them because they were marked as Reviewed on Notion, but noticed this once I checked the preview. Looking at Crowdin, the `json` file for all three of these languages is 83%. Perhaps we jumped the gun on them being ready for a PR? cc: @lukassim

samajammin (approved):
Chatted offline - pulling in 👍

PR #466, opened 2022-03-09
Add Erigon, remove OE 

timbeiko:
This PR adds information about Erigon to the launchpad and, at the same time, removes OpenEthereum which has been deprecated ([announcement](https://medium.com/openethereum/gnosis-joins-erigon-formerly-turbo-geth-to-release-next-gen-ethereum-client-c6708dd06dd)). 

Combining both these changes in a single means we don't have to deal with the launchpad showing 5 clients for a period of time. That said, the OE repo seems to have activity and the client may be maintained by another team, so it's really a "quick fix" as we could potentially have to add it again if/when they support The Merge. The changes have been made in separate commits, which can be cherry-picked if we'd rather split them. 

Note: I wasn't sure of the best way to remove all translations for the OE documentation, so I did not do it.


hwwhww (approved):
I pushed commits for
1. Fix `ClientId`
2. Update translations
3. Minor changes


LGTM!

PR #467, opened 2022-03-10
Add Kintsugi and Kiln testnets

hwwhww:
Target branch: `bellatrix`

- Add Kintsugi and Kiln configs
- Make it feasible on Metamask
- Disable Fortmatic and Portis for now

TODOs:
- [x] make deposit on kintsugi
- [ ] test top-up
- [x] test Kiln
    - note that beaconchain explorer doesn't support APIs before genesis. some beaconcha.in-API calls fail now


hwwhww:
Merging this PR now to see the demo website. Post-merge feedback is still appreciated!

hwwhww:
Hi @samajammin @wackerow

Could you help review this PR? Especially the security check of https://github.com/ethereum/staking-launchpad/blob/bellatrix/src/pages/ConnectWallet/Chains.ts?

samajammin:
Hey @hwwhww! I'm OOO for the rest of the week. Perhaps @wackerow can help take a look. Otherwise I'll check this out next week when I'm back.

wackerow (commented):
Hey @hwwhww, this looks pretty good. Only thought I had on this would be to pass the web3provider already available in the `WalletButton.tsx` file when the `changeToTestnet` function is called. We could adjust that function (only used there) to accept a provider as a prop, then we wouldn't have to interact with the `window` object.

That being said, don't think this would cause any security issue. We use a similar pattern on the CLR fund page, except we don't call the `wallet_switchEthereumChain` for error handling. Not sure if this is needed, since fairly sure the `wallet_addEthereumChain` will trigger a switch if the user already has it installed, and if MetaMask is not available this method will also not work. 

Again, don't see anything critical here. Hope that helps.

PR #461, opened 2022-02-25
Adding Wagyu Key Gen option to the generating keys section

remyroy:
This PR adds the Wagyu Key Gen option to the generating keys section. [Wagyu Key Gen](https://github.com/stake-house/wagyu-key-gen) is a GUI wrapper on top of staking-deposit-cli (previously known as eth2.0-deposit-cli) to generate your keys. It was fully audited. It works well on Linux, Windows and Mac.

Closes #460

PR #470, opened 2022-03-13
Reorder keygen options

CarlBeek:
## Motivation
In #461, we added Wagyu GUI as a new method for key generation. Very cool and thanks for the work, @remyroy, @wackerow, and the Steak-House team!

I've spent a few hours going through the https://github.com/stake-house/wagyu-key-gen codebase and it passes the sniff test. That said, I am not quite comfortable offering Wagyu as the default offering on mainnet **yet**.

## What this PR actually does

So, this PR sets the CLI as the default offering when on `mainnet`, but when using testnets the default becomes Wagyu. This way we can build confidence in Wagyu over time for a bit and in a few months, update to GUI first. :)


wackerow (approved):
Nice, good call @CarlBeek, this looks good. Agree this is a safer approach to roll this out over time, and this is what we have the testnets for.

Just updated one small JS adjustment, otherwise looks good!

PR #485, opened 2022-05-17
Use English "Goerli" instead of typo German "Göerli"

hwwhww:
1. Using German is config made us have to use `NetworkChainId['Göerli']`. As we are adding Ropsten and other EL testnets, it would be easier in parameterization if we can just use English "Goerli".
2. It turns out `Göerli` is a typo and it should have been `Görli`. 😅  Let's just use `Goerli`.


CarlBeek (approved):
Good catch. Indeed, it seems simpler to unpack the umlaut here 👍

PR #487, opened 2022-05-18
FAQ link patch

wackerow:
## Description
Adds `inline` and `primary` props where needed to FAQ links

## Issue:
None filed, but note example here:
image

With fix:
image


hwwhww (approved):
Thanks 👍

PR #495, opened 2022-05-27
Add top-up modal escape [Fixes #383]

### LINKED ISSUE #383: Top Up page modal has no escape
Top Up page presents a modal that has no close button and clicking outside the modal does not close it. 

Steps to reproduce:
1. Go to https://pyrmont.launchpad.ethereum.org/en/top-up or https://launchpad.ethereum.org/en/top-up
2. See modal pop-up, but cannot click anything in background to go back.
(More natural flow would be to click on the Top Up button on the page, which would still allow a user to navigate "Back", but a direct link leaves you completely stuck)

Suggestion
1. Allow user to click outside the Top Up modal area to return to page before Top Up. 
2. And/or make a non-modal Top Up page, with a modal that can be launched from that (instead of popping up directly from the nav button)

wackerow:
## Description
- Adds closing `X` to modal pop-up on `/top-up` page
- Adds "Connect wallet" button to same page if user has not connected, and has exited the modal

## Related issue
[Fixes #383]

### DIFF
--- a/src/pages/TopUp/components/WalletConnectModal.tsx
+++ b/src/pages/TopUp/components/WalletConnectModal.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo, useState } from 'react';
 import { FormattedMessage } from 'react-intl';
+import styled from 'styled-components';
 import { useWeb3React } from '@web3-react/core';
 import { Web3Provider } from '@ethersproject/providers';
 import { AbstractConnector } from '@web3-react/abstract-connector';
@@ -15,6 +16,7 @@ import {
 import { WalletButton } from '../../ConnectWallet/WalletButton';
 import { web3ReactInterface } from '../../ConnectWallet';
 import metamaskLogo from '../../../static/metamask.svg';
+import closeGlyph from '../../../static/close.svg';
 import {
   ENABLE_RPC_FEATURES,
   IS_MAINNET,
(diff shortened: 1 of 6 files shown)


hwwhww:
Just commenting on the preview page, it looks great! 👍 

Defer to @samajammin on the TypeScript magic.

samajammin (changes requested):
Nice @wackerow!

This may have already been an issue (i.e. unrelated from your PR) but I noticed when I'm on a different network other than Mainnet (e.g. Arbitrum), it creates some funky behaviour when trying to connect (nothing happens when I click MetaMask). If I'm on Mainnet, it correctly tells me to switch to Goerli.

Mind making a change here to check if user is on ANY network other than the correct network? If so, we should display that modal to switch:
Screen Shot 2022-05-30 at 9 33 47 AM


If you'd like, feel free to create a separate issue for this... ideal logic here might be to auto-switch the network for MetaMask users.

src/pages/TopUp/components/WalletConnectModal.tsx
```diff
@@ -47,10 +67,14 @@ const WalletConnectModal: React.FC = () => {
     return !Object.values(AllowedNetworks).includes(network);
   }, [chainId]);
 
+  const handleClose = () => {
+    setLoading(false);
+  };
```
pettinarip, line 72: I'm ok with it. But a nice to have:

Separate concerns a bit here:
- Lets control the `loading` state at the consumer level and pass only that state to the component.

```tsx
// at the consumer level

const handleWalletModalClose = () => {
  setLoading(false);
};

return <WalletConnectModal loading={loading} onClose={handleWalletModalClose} />

```

pettinarip (approved):
LGTM! left a couple of small suggestions, nothing blocking.

src/pages/TopUp/components/WalletConnectModal.tsx
wackerow: Pushed a patch for this... removed the need to rename the above `_`

wackerow:
@samajammin Yeah this is another existing issue it looks like. Opened an issue for it, but gonna hold on building it into this PR. Can revisit separately.

PR #505, opened 2022-06-09
Fix typo 'there will only be Ethereum'

JonAndrews:


wackerow (approved):
Thanks @JonAndrews! Good catch =)

PR #520, opened 2022-07-13
Add translation support for nested strings

wackerow:
Fixes issue where intl variables used within `FormatMessage` component were not always set up for translation themselves.

example:
```ts
// Incorrect, as "not" does not get translated
              values={{ not: <em>not</em> }}
// Corrected to:
              values={{
                not: (
                  <em>
                    <FormattedMessage defaultMessage="not" />
                  </em>
                ),
              }}
```

- Extracts remaining strings, including "Mainnet" and "____ testnet" for translation.
- Also capitalized "Mainnet" and the testnet names for consistency

Note: There are some CLI variables such as `--num_validators` and `--chain` that are currently being translated. @CarlBeek Want to confirm that this is aligned with the `staking-deposit-cli` script.

### DIFF
--- a/src/components/AppBar.tsx
+++ b/src/components/AppBar.tsx
@@ -8,7 +8,6 @@ import { useWeb3React } from '@web3-react/core';
 import { Web3Provider } from '@ethersproject/providers';
 import EthDiamond from '../static/eth-diamond-plain.svg';
 import { web3ReactInterface } from '../pages/ConnectWallet';
-import { trimString } from '../utils/trimString';
 import {
   AllowedNetworks,
   NetworkChainId,
@@ -24,8 +23,9 @@ import {
   MAINNET_LAUNCHPAD_URL,
   TESTNET_LAUNCHPAD_NAME,
   TESTNET_LAUNCHPAD_URL,
-  EL_TESTNET_NAME,
 } from '../utils/envVars';
+import { trimString } from '../utils/trimString';
+import useNetworkName from '../hooks/useIntlNetworkName';
 import useMobileCheck from '../hooks/useMobileCheck';
 
 const RainbowBackground = styled(Box)`
@@ -124,6 +124,15 @@ const _AppBar = ({ location }: RouteComponentProps) => {
     account,
     chainId,
   }: web3ReactInterface = useWeb3React<Web3Provider>();
+  const { executionLayerName, consensusLayerName } = useNetworkName();
+  const oppositeNetwork = IS_MAINNET ? (
+    <FormattedMessage
+      defaultMessage="{TESTNET_LAUNCHPAD_NAME} testnet"
+      values={{ TESTNET_LAUNCHPAD_NAME }}
+    />
+  ) : (
+    <FormattedMessage defaultMessage="Mainnet" />
+  );
 
   let network;
   let networkAllowed = false;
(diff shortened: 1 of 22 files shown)


src/components/AppBar.tsx
```diff
@@ -151,9 +151,15 @@ const _AppBar = ({ location }: RouteComponentProps) => {
     ? TESTNET_LAUNCHPAD_URL
     : MAINNET_LAUNCHPAD_URL;
 
-  const networkName = IS_MAINNET
-    ? 'mainnet'
-    : `${EL_TESTNET_NAME.toLowerCase()} testnet`;
+  const networkName = IS_MAINNET ? (
```
samajammin, line 154: Any particular reason we create a variable here vs. placing this inline with the HTML like we do elsewhere?

src/pages/TopUp/components/WalletConnectModal.tsx
```diff
@@ -88,11 +88,16 @@ const WalletConnectModal: React.FC<{
             <FormattedMessage
               defaultMessage="Connect to {network}"
               values={{
-                network: IS_MAINNET
-                  ? 'Ethereum mainnet'
-                  : `${EL_TESTNET_NAME.toLowerCase()} testnet`,
+                network: IS_MAINNET ? (
```
samajammin, line 91: Seems like we duplicate this exact same logic in a few places - I wonder if we could make this more DRY?

samajammin (approved):
A couple minor suggestions - overall looks good!

src/pages/TopUp/components/WalletConnectModal.tsx
wackerow: Agree with this sentiment... Just added some commits implementing a `useIntlNetworkName` hook to remove a lot of the redundancy here.

Can access using `const { executionLayerName, consensusLayerName } = useIntlNetworkName();`

src/hooks/useIntlNetworkName.tsx
```diff
@@ -0,0 +1,29 @@
+import { useIntl } from 'react-intl';
+import {
+  IS_MAINNET,
+  EL_TESTNET_NAME,
+  TESTNET_LAUNCHPAD_NAME,
+} from '../utils/envVars';
+
+const useIntlNetworkName = (): {
```
samajammin, line 8: 🔥

samajammin:
I thought the code looks good but something appears to be screwy with the logic:
Image 2022-07-27 at 11 07 15 AM
- lowercase `mainnet` should be uppercase
- Either the current networks is wrong or the `Switch to Mainnet` is wrong

samajammin:
I'm still confused. Why would we use `NETWORK_NAME` = `'Mainnet'` if `IS_MAINNET` is `false`?

Perhaps we should update:
```
export const NETWORK_NAME          = (IS_MAINNET || !process.env.REACT_APP_TESTNET_LAUNCHPAD_NAME) ? 'Mainnet' : process.env.REACT_APP_TESTNET_LAUNCHPAD_NAME;
```
to be
```
export const NETWORK_NAME          = IS_MAINNET ? 'Mainnet' : TESTNET_LAUNCHPAD_NAME;
```
Given that we have a fallback for that variable if `REACT_APP_TESTNET_LAUNCHPAD_NAME` is falsey?

samajammin (approved):
👍

PR #539, opened 2022-08-23
Add protection to mainnet deposit contract address

hwwhww:
Fix deposit contract address for mainnet.

@wackerow I copied your `MergeNotification` object to make a `TestnetNotification`. :p


CarlBeek (approved):
This is definitely a security improvement for our deployments and is a good sanity check for users running this locally.

samajammin:
@wackerow any idea why the preview deploy appears to be broken?
https://deploy-preview-539--serene-carson-3331d5.netlify.app/

Seems this is also the case on #535:
https://deploy-preview-535--serene-carson-3331d5.netlify.app/

Missing ENV variables in Netlify perhaps?

wackerow:
@samajammin I'm not sure we're getting the same results. The preview deploys are loading fine for me... maybe clear some cache or try another browser?

Only thing I'm noting is the Goerli metrics aren't being pulled in (which appears to also be the case on the live page... likely related to the merge and name change, I can file a bug report on that)

src/components/TestnetNotification.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Alert } from './Alert';
+
+const StyledAlert = styled(Alert)`
+  text-align: center;
+`;
+
+export const TestnetNotification = (): JSX.Element => {
+  return (
+    <StyledAlert variant="warning" round="none" pad="small">
+      <FormattedMessage defaultMessage="Warning: This is a staking launchpad for TESTNET." />
```
wackerow, line 13: Slight wording suggestion:
```suggestion
      <FormattedMessage defaultMessage="Warning: This is a staking launchpad for the TESTNET." />
```

Or, we could pull the all-caps styling into css so it doesn't get goofed up in translation:
```suggestion
      <FormattedMessage
        defaultMessage="Warning: This is a staking launchpad for {testnet}."
        values={{ testnet: <AllCaps><FormattedMessage defaultMessage="testnet" /></AllCaps> }}
      />
```
(Plus the `const AllCaps` suggestion above)

src/components/TestnetNotification.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import styled from 'styled-components';
+import { Alert } from './Alert';
+
+const StyledAlert = styled(Alert)`
+  text-align: center;
+`;
+
```
wackerow, line 9: (If going with the suggestion below)
```suggestion

const AllCaps = styled.span`
  text-transform: uppercase;
`;

```

wackerow (commented):
@hwwhww Thanks for doing this 🎉  Left a couple suggestions related to styling and trying to limit any issues with translations. Let me know if they make sense =)

samajammin:
> The preview deploys are loading fine for me

Hmmm thanks for checking @wackerow. They do load fine in other browsers & in incognito but for some reason is failing for me in Brave, despite clearing cache/storage/cookies:
Image 2022-08-23 at 1 07 54 PM

In any case, the issue is clearly on me end, so feel free to disregard 😄

samajammin (commented):
LGTM (with @wackerow's suggested changes)!

hwwhww:
@samajammin @wackerow I also have the same problem with the non-incognito Brave browser 😭

PR #547, opened 2022-08-25
Update Lighthouse link

paulhauner:
Heyo :wave: 

The Lighthouse blog has moved from `lighthouse.sigmaprime.io` to `lighthouse-blog.sigmaprime.io`.

I believe this PR makes the appropriate change, however I'm not so familiar with this code-base so some extra eyes would be appreciated :pray:


wackerow (approved):
Thanks for catching this @paulhauner!

PR #557, opened 2022-09-23
Deprecate MergeReadiness

wackerow:
- Removes notifications related to The Merge
- Removes MergeReadiness checklist page
- Moves pertinent information from this page to corresponding places in the existing FAQ and Checklist pages
- This includes information related to setting up a JWT and fee recipient
- Adds client-diversity notice to Lighthouse page
- Updates copy to be correct in a post-merge environment
- Updates notes about Shanghai to inform that withdrawal functionality is _proposed_ to be included in this upgrade. Removing stating inclusion as fact until further decisions have been made regarding this upgrade.


wackerow:
Added a few commits to clean up some of the links... original used the name of the document from each client docs page like "Configure Fee Recipient" but this ended up being redundant, and would just add to the number of strings that need to be translated.

I removed these labels, pulled any client specific links into the clientInfo object, and render these links as just pipe-separated inline lists, adding some text in front of each to identify the type of doc.

src/pages/Checklist/index.tsx
```diff
@@ -483,6 +513,11 @@ export const Checklist = () => {
                 />
               </Text>
             </li>
+            <li className="py5">
+              <Text>
+                <FormattedMessage defaultMessage="At time of the Merge, third-party providers (such as Infura and Alchemy) will no longer be viable options to outsource execution layer responsibilities. All stakers must run a pair of both an execution and consensus client to properly attest to the network." />
```
hwwhww, line 518: Should we use the present tense for the post-merge scenario?
```suggestion
                <FormattedMessage defaultMessage="After the Merge, third-party providers (such as Infura and Alchemy) are no longer viable options to outsource execution layer responsibilities. All stakers must run a pair of both an execution and consensus client to properly attest to the network." />
```

src/pages/FAQ/index.jsx
```diff
@@ -159,6 +159,14 @@ export const FAQ = () => {
                 }}
               />
             </Heading>
+            <Text className="mt20 mb20">
+              <FormattedMessage
+                defaultMessage="Each 32 ETH deposit activates one set of validator keys. These keys are used to sign off on the state of the network. The lower the ETH requirement, the more resulting signatures must be saved by the network. {pricePerValidator} ETH was chosen as a balance between enabling as many people as possible to stake without inhibiting decentralization by bloating the size of each block with signatures."
```
hwwhww, line 164:
```suggestion
                defaultMessage="Each {pricePerValidator} ETH deposit activates one set of validator keys. These keys are used to sign off on the state of the network. The lower the ETH requirement, the more resulting signatures must be saved by the network. {pricePerValidator} ETH was chosen as a balance between enabling as many people as possible to stake without inhibiting decentralization by bloating the size of each block with signatures."
```

src/pages/FAQ/index.jsx
```diff
@@ -232,12 +237,7 @@ export const FAQ = () => {
               <FormattedMessage defaultMessage="Can I stop running my validator for a few days and then start it back up again?" />
             </Heading>
             <Text className="mt10">
-              <FormattedMessage
-                defaultMessage="If you go offline for a number of days under normal conditions you will lose an amount of ETH
-                  roughly equivalent to the amount of ETH you would have gained in
-                  that period. In other words, if you stood to earn ≈0.01 ETH, you
-                  would instead be penalised ≈0.01 ETH."
-              />
+              <FormattedMessage defaultMessage="Yes, but with small penalties. If you go offline for a number of days under normal conditions you will lose an amount of ETH roughly equivalent to the amount of ETH you would have gained in that period. In other words, if you stood to earn ≈0.01 ETH, you would instead be penalised ≈0.01 ETH." />
```
hwwhww, line 240:
```suggestion
              <FormattedMessage defaultMessage="Yes, but with small penalties. If you go offline for a number of days under normal conditions you will lose an amount of ETH roughly equivalent to the amount of ETH you would have gained in that period. In other words, if you stood to earn ≈0.01 ETH, you would instead be penalized ≈0.01 ETH." />
```

hwwhww (approved):
Great work! It's not just deprecating the merge checklist but also some thoughtful proofreading work. 👍

src/pages/Checklist/index.tsx
jmcook1186: alternatively, just to make it very clear we are alreayd post-merge...
```suggestion
                <FormattedMessage defaultMessage="Since the Merge, third-party providers (such as Infura and Alchemy) are no longer being viable options to outsource execution layer responsibilities. All stakers must run a pair of both an execution and consensus client to properly attest to the network." />
```

wackerow: Good catch

PR #564, opened 2022-10-21
Import latest translations from Crowdin

wackerow:
### Description
Imports latest translated content from Crowdin for all languages supported on the site


CarlBeek (approved):
Based off of some random spot-checks, LGTM

PR #575, opened 2022-12-14
fix: use correct hover state for disabled tile

dzbo:
#### Description

Atm when one of tiles is disabled it still show `pointer` cursor while hovering on the text part.

#### Examples

Before fix:
CleanShot 2022-12-14 at 17 02 27

After fix:
CleanShot 2022-12-14 at 17 03 32


wackerow (approved):
LGTM, thanks @dzbo!

This shines light on some other problems with those "buttons"... like how they aren't actually buttons, and they should be (lacking any a11y features of an actual button). No need for scope creep in this PR though, will bring it in. Thanks!

PR #541, opened 2022-08-23
Adds Lodestar support

CarlBeek:
Now that Lodestar [has released](https://github.com/ChainSafe/lodestar/releases/tag/v1.0.0) `v1.0.0`, this PR re-implements #436 just on-top of a fresher branch of `dev`.

PR #586, opened 2023-02-03
`dev` -> `master` release

wackerow:
### Change log
- #474 Update messaging on withdrawals
- #541 Adds Lodestar support
- #571 Patch: Teku JWT docs link
- #572 Bump decode-uri-component from 0.2.0 to 0.2.2
- #573 Bump qs from 6.5.2 to 6.5.3
- #574 Bump express from 4.17.1 to 4.18.2
- #575 fix: use correct hover state for disabled tile
- #581 Bump ua-parser-js from 0.7.31 to 0.7.33


hwwhww (approved):
LGTM!

one not-strong nitpicking :
On my laptop, the CL client layout looks like this in a full-size view:
image

It looks better in a more narrow window via RWD:   
image

It would be nice if it can be auto-adjusted to "first row: 3 clients, second row: 2 clients" automatically in the full-size view.

@wackerow @samajammin  feel free to go merge it. :shipit:

wackerow:
@hwwhww Agree! I'll patch this in a separate PR and we can add it in after this, thanks for pointing it out =)

PR #588, opened 2023-02-07
Wrap client cards early if over four count

wackerow:
## Description
- Evens out the visual display of the 4 EL clients and 5 CL clients on the select client pages
- 4 cards fit nicely on desktop, but 5 causes one to wrap. This add logic to put a max-width if there are >4 clients (CL clients) so it wraps after the third card instead

Noted in #586


hwwhww (approved):
lgtm! thank you @wackerow for ironing out my nitpicking 👍

PR #592, opened 2023-02-08
Fix top-up where the EL address has only made one deposit 

hwwhww:
It was reported by @barnabasbusa.

### Description
The case is: if the user has only made **one** deposit before, our API response phaser would break. But if the user has made multiple deposits, it's alright.

Test case:
- The EL address has made only one deposit: https://zhejiang.beaconcha.in/api/v1/validator/eth1/0xF9dBc5fc75b5802916C66dae672e0aa124EDcD3e
    - Error: it would go all-blank page once I connect the wallet. 
    - `validators.map` TypeError: a.map is not a function at ValidatorTable.tsx:123:23
- The EL address has made multiple deposits: https://zhejiang.beaconcha.in/api/v1/validator/eth1/0xED6ee41b4Eee71Cc179683C3b90eFe20A1744D44
    - This is all good. 

### How did I fix it
- Rename `BeaconChainValidatorResponse` to `BeaconChainValidatorEth1Deposit` because it was really confusing with `BeaconChainValidator`
- Accept that `data` might be a `BeaconChainValidator` Object rather than a `BeaconChainValidator` Array. And check it with `const response = Array.isArray(data) ? data : [data];`

I fixed it but I'm not really sure why we need it. I simply copied the pattern from the `BeaconChainValidatorEth1Deposit` API call. @wackerow could you help review it? 🙏

### DIFF
--- a/src/pages/TopUp/index.tsx
+++ b/src/pages/TopUp/index.tsx
@@ -7,7 +7,7 @@ import { FormPrevious } from 'grommet-icons';
 import { Alert as AlertIcon } from 'grommet-icons/icons';
 import {
   BeaconChainValidator,
-  BeaconChainValidatorResponse,
+  BeaconChainValidatorEth1Deposit,
   Props,
 } from './types';
 import { Text } from '../../components/Text';
@@ -105,7 +105,9 @@ const _TopUpPage: React.FC<Props> = () => {
           ({
             data,
           }: {
-            data: BeaconChainValidatorResponse[] | BeaconChainValidatorResponse;
+            data:
+              | BeaconChainValidatorEth1Deposit[]
+              | BeaconChainValidatorEth1Deposit;
           }) => {
             setShowDepositVerificationWarning(false);
             const response = Array.isArray(data) ? data : [data];
@@ -114,11 +116,6 @@ const _TopUpPage: React.FC<Props> = () => {
               setValidators([]);
               setLoading(false);
               return;
-            }
-
-            if (response.length === 0) {
-              setValidators([]);
-              setLoading(false);
             } else {
               // query by public keys
               const pubKeysCommaDelineated = `${response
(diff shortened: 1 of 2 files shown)


wackerow:
@hwwhww I'll take a look... Both URLs you noted return an object with _array_ values for the `data` property, so I'm not sure why we would need to account for this.

src/pages/TopUp/index.tsx
```diff
@@ -130,13 +127,20 @@ const _TopUpPage: React.FC<Props> = () => {
                 `${BEACONCHAIN_URL}/api/v1/validator/${pubKeysCommaDelineated}`
               )
                 .then(r => r.json())
-                .then(({ data }: { data: BeaconChainValidator[] }) => {
-                  if (data.length === 0) {
-                    setShowDepositVerificationWarning(true);
+                .then(
```
wackerow, line 130: @hwwhww Looks like this is actually the only response that has the Array vs Object caveat... not the `eth1` fetch above. This code could use a little modernizing as well, so I can try to clean this up and push the results if you'd like.

src/pages/TopUp/index.tsx
```diff
@@ -114,11 +116,6 @@ const _TopUpPage: React.FC<Props> = () => {
...
               setLoading(false);
               return;
-            }
-
-            if (response.length === 0) {
-              setValidators([]);
-              setLoading(false);
             } else {
               // query by public keys
               const pubKeysCommaDelineated = `${response
```
wackerow, line 121: Side note, is there a reason we're chopping this list at 50 items? Is this due to limitations on the beaconcha.in api call below?

wackerow:
Realizing this was targeting the `shapella` branch, as these changes aren't on `dev`. @hwwhww Is there a plan to merge `shapella` back into `dev`?

PR #593, opened 2023-02-08
Top up data fetching refactor

wackerow:
Merge requested: `fix-top-up-one-deposit` <- `fix-top-up-one-deposit-refactor`

Following up on @hwwhww's PR #592, realized we could do a fair amount to clean up the fetching logic being performed on this component. Given I overhauled this function, figured I would post as a separate PR instead of as a suggestion on #592, or just committing directly to that branch.

- Switches to async/await syntax
- Uses `new URL()` objects to build dependable URL link
- Uses try/catch/finally block with guard clauses for error handling

These changes help remove the function chaining and nesting of code, and should improve readability and reliability.

Additionally, this cleans up the expected return values for the data we're fetching.
- The first call is fetching all pubKeys for a given eth1 address, and _always_ returns type `BeaconChainValidatorEth1Deposit[]` (an array)
- The second call fetches validator details by pub key... if only one pub key is provided, an object is returned, but if a comma-separated list of pub keys is passed, the response is an array... thus its type would be `BeaconChainValidator | BeaconChainValidator[]`

### DIFF
--- a/src/pages/TopUp/index.tsx
+++ b/src/pages/TopUp/index.tsx
@@ -112,7 +112,6 @@ const _TopUpPage: React.FC<Props> = () => {
         // No validators for that user's wallet address, exit early
         if (dataEth1.length === 0) {
           setValidators([]);
-          setLoading(false);
           return;
         }
         // Prepare comma delimited list of pubKeys for eth1 address provided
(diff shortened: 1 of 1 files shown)


hwwhww (approved):
Not a JS asynchronous expert but I think `try/catch/finally` is much more readable than multiple `fetch`. ❤️ I tested the 3 cases (0 deposit, 1 deposit, and multiple deposits) with this branch and they are all good. Thank you @wackerow!

wackerow:
Thanks for the review @hwwhww! This PR was against your #592 PR, so given your approval here I'll go ahead and merge that to your branch to close this out.

PR #594, opened 2023-02-08
Update repo for RTL support

wackerow:
## Description
We currently support multiple languages on the repo, but Arabic is a right-to-left language that does not yet have proper support on this site. The text is translated, but the page layout does not match what RTL languages are used to.

This implemented the `dir="rtl"` attribute on the `html` tag for Arabic. In doing so, the layout of all DOM items switches from LTR to RTL.

In doing so, we needed to update any usage of "left" and "right" as absolute directions. For example:
- `margin-left` will ALWAYS place the margin on the left, but `margin-inline-start` will place the margin at the beginning of the text (which will be left for LTR, or automatically switch to right for RTL)
- Usage of "gap" where possible to space children without having to deal with margins at all, which also helps when flex items collapse to a stack (gap still applied vertically)
- Absolute positioning of "left" or "right" should be "inset-inline-start" or "inset-inline-end"

All CSS properties were updated to accommodate for these changes. The base `GlobalStyles` file was also updated to switch any further use of `[ml/mr/pl/pr]` to use "inline-start/end" options instead of "left" or "right"

Added some custom logic to flip the external arrow on links, and the chevrons in the form flow and checklist ToC items. 

## Preview link
https://deploy-preview-594--serene-carson-3331d5.netlify.app/ar

## Screenshot preview
image
image
image


hwwhww (approved):
good point!!!

It seems some command line might need some special treatment.
EL client page:
image

`generate-keys` page:
image

Honestly, I don't know what's the presentation like on Arabic users' terminals. 😅

wackerow:
@hwwhww Good catch! Those commands should not be getting translated in the first place, since those need to be very specific inputs for the CLI. I'll form a patch for that

wackerow:
Updated with a couple things:
- `TRANSLATE_CLI_FLAGS` env var boolean, defaulting to FALSE. Used this to make all `--` flags _conditionally_ translated only if that env var is set to TRUE, otherwise it will default to using English strings for technical commands.
- With all "code" being in English, set `direction: ltr;` on all pre-formatted code blocks to render these items in LTR

image

Input from a native speaker would certainly be helpful here, but I would imagine that this makes the most sense... It may look a little off going back and forth, but it at least frames the different text in their appropriately alignments where possible.

wackerow:
FWIW, checked with a friend who is a native speaker and said this approach appears correct.

PR #602, opened 2023-02-14
Update data fetching from Beaconcha.in [Fixes #597]

### LINKED ISSUE #597: Unify topline metrics across e.org and Launchpad
For a more detailed description, please reference:
https://github.com/ethereum/ethereum-org-website/issues/9232

To cut to the chase, the topline metrics at Ethereum.org vs the Launchpad are calculated differently and give different results, which causes confusion.

image

After investigating, @wackerow determined that the following would be appropriate:

"Again, I think we should update e.org to use the "latest epoch" endpoint matching the Launchpad, and I think we should open this as an issue on the Staking Launchpad repo to update the endpoint for "Total ETH staked" and "Current APR" to match that of ethereum.org."

My only (personal) suggestion is: it would be nice / intuitive if whichever amounts are displayed for "Total ETH stake" & "validator count" have the property that multiplying "validator count" by 32 = "Total ETH stake". Ethereum.org currently has this property (though it should be rounding to an integer = bug). If this property doesn't hold in the revised version, maybe place a comment nearby explaining how this calc is done / the logic behind it? Just to avoid any confusion.

Thanks,

wackerow:
## Description
Updates the APR and Total ETH Staked metrics to pull from [Beaconcha.in's ETH.STORE](https://beaconcha.in/ethstore) for more real-time and accurate APR, and total cumulative effective balance, without need to calculate an estimate. The total number of validators still needs to be fetched from the existing `epoch/latest` endpoint.

- Adds a `fetch` utility file for each data endpoint
- Adds a reusable hook that fetches the data from both endpoints, and returns a `state` object with these three metrics
- This uses `Intl.NumberFormat(local, {options})` to format the numbers accordingly.
- Implements this hook on both landing page and Congratulations page where metrics are shown
- This keeps the "total validators" count the same as it was, but updates the "total ETH staked" and "apr" to be more accurate, and to match with what Beaconcha.in and ethereum.org/staking display.
- Removes deprecated `queryBeaconchain`, `numberWithCommas` and `calculateStakingRewards` utility functions

(Note, `Intl.NumberFormat(locale).format(value)` should be used instead of `numberWithCommas` custom formatting for improved i18n support)

## Related issue
- Fixes #597


wackerow:
Parallel PR open at https://github.com/ethereum/ethereum-org-website/pull/9458

src/hooks/useBeaconchainData.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useEffect(() => {
+    // Number formatters
+    const formatPercent = (value: number): string =>
+      new Intl.NumberFormat(locale, {
+        style: 'percent',
+        minimumSignificantDigits: 3,
+        maximumSignificantDigits: 3,
+      }).format(value);
+    const formatNumber = (value: number): string =>
+      new Intl.NumberFormat(locale).format(value);
```
pettinarip, line 33: Would suggest moving these functions to the `utils` function. Not a blocker though

src/hooks/useBeaconchainData.tsx
```diff
@@ -0,0 +1,64 @@
...
+        statusCode,
+      }: FetchTotalValidatorsResponse = await fetchTotalValidators();
+      setState(prev => ({
+        ...prev,
+        totalValidators: formatNumber(totalValidators),
+        status: Math.max(statusCode, prev.status),
+      }));
+    })();
+  }, [locale]);
+  return { state };
```
pettinarip, line 63: nitpick: return just the `state` object. No real need to have it nested. From the consumer side you could just spread and pick whatever you like:

`const { apr } = useBeaconchainData()`
or `const data = useBeaconchainData()`
```suggestion
  return state;
```

src/utils/fetchTotalStakeAndAPR.ts
```diff
@@ -0,0 +1,41 @@
...
+    apr: number;
+    msg?: unknown;
+  };
+}
+export const fetchTotalStakeAndAPR = async (): Promise<FetchTotalStakeAndAPRResponse> => {
+  const { href } = new URL('/api/v1/ethstore/latest', BEACONCHAIN_URL);
+  try {
+    const response: AxiosResponse = await axios.get(href);
+    if (response.status !== 200) throw new Error(response.statusText);
+    const { data }: { data: EthstoreData } = response;
```
pettinarip, line 24: These types are kind of weird. I think this would simplify it a bit
```suggestion
    // you can pass the data type on the `get` function
    const response = await axios.get<EthstoreData>(href);
    if (response.status !== 200) throw new Error(response.statusText);
    const { data } = response;
```

src/utils/fetchTotalValidators.ts
```diff
@@ -0,0 +1,36 @@
...
+    msg?: unknown;
+  };
+}
+
+export const fetchTotalValidators = async (): Promise<FetchTotalValidatorsResponse> => {
+  const { href } = new URL('/api/v1/epoch/latest', BEACONCHAIN_URL);
+  try {
+    const response: AxiosResponse = await axios.get(href);
+    if (response.status !== 200) throw new Error(response.statusText);
+    const { data }: { data: EpochData } = response;
```
pettinarip, line 22: same here with the types around axios

pettinarip (approved):
Small suggestions but overall this looks good @wackerow 💯

src/hooks/useBeaconchainData.tsx
wackerow: Good call... built it this way to match the existing `const { state }` destructuring that was being used, but agree we can simplify.

src/utils/fetchTotalValidators.ts
wackerow: Nice, thanks for suggesting this...  much cleaner

src/hooks/useBeaconchainData.tsx
wackerow: Good call... built it this way to match the existing `const { state }` destructuring that was being used, but agree we can simplify.

PR #620, opened 2023-04-01
Update to "Command line with flags" example

calumtomeny:
The example includes the "0x" prefix when specifying the withdrawal credentials - the "0x" prefix should not be included. If the "0x" prefix is included in the credentials list then the user will see an error. Fixes issues #621.


wackerow:
Left a comment on the issue (#621)

wackerow:
Discussion thread on issue #621, closing this out

PR #623, opened 2023-04-03
Add missing word (blocks)

sandakersmann:
Add missing word (blocks)


wackerow (approved):
Thanks @sandakersmann! Good catch

PR #626, opened 2023-04-04
Refactor Code component to fix direction issues

wackerow:
## Description
- Moves hard-coded LTR direction assignment to an inner-span that only styles the text itself, allowing the surrounding span to be styled with direction-responsive margins and padding.
- Patches one use case of a `span` selector used over the Code component, that should only select the outer span


pettinarip (approved):
🚀

PR #633, opened 2023-04-18
Import latest translations from Crowdin

wackerow:
## Description
- Imports latest translations from Crowdin


pettinarip (changes requested):
In https://deploy-preview-633--serene-carson-3331d5.netlify.app/zh/btec I don't see the first paragraph being translated. In [prod we have the translations](https://launchpad.ethereum.org/zh/btec).

Is it possible that this branch is not updated? or did we miss something?

PR #666, opened 2023-11-24
update Nimbus links

narimiran:


src/pages/Clients/Consensus/Nimbus.tsx
```diff
@@ -21,8 +21,8 @@ export const NimbusDetails = ({ shortened }: { shortened?: boolean }) => (
       <FormattedMessage
         defaultMessage="Nimbus is a client implementation for both execution and consensus layers
                         that strives to be as lightweight as possible in terms of resources used.
-                        This allows it to perform well on embedded systems, resource-restricted devices --
-                        including Raspberry Pis -- and multi-purpose servers."
+                        This allows it to perform well on embedded systems, resource-restricted devices —
+                        including Raspberry Pis — and multi-purpose servers."
```
narimiran, line 25: In code-font it's not visible/obvious, but this is now changed to an em-dash: — , so it is rendered correctly, since two hyphens (--) were not rendered as an em-dash.

wackerow, line 25: Thanks @narimiran! One step further, we should remove the spaces as well... will update =)

wackerow (approved):

src/pages/Clients/Consensus/Nimbus.tsx
narimiran: AFAIK, that's just an American rule about no spaces around em-dash, it is not universal.

PR #673, opened 2023-12-13
Updated Nimbus background image and reference

liftlines:
Updated Nimbus background image on the Clients page.


liftlines:
hi @hwwhww can you please help review and merge? Nixo (ethstaker) suggested I contact you as there are no reviewers identified.

zah:
Perhaps @wackerow could look into this as well?

wackerow (approved):
Hey @liftlines, thanks for posting!

Just took a look... sounds like this is a branding update where we don't want to use the old anymore.

I went ahead and just removed the old image assets and renamed the new ones to match to keep it consistent. There was a stray/duplicate/unused image in the `/components` folder which I removed. Simplifies the diff, and this should do it!

wackerow:
@liftlines One problem, this PR needs to target `dev` branch instead of `master`... I tried updating but GitHub is blocking me.. any chance you could do that? Might require opening this branch as a new PR... let me know if you have issues, I can open a separate one if needed.

Issue #549, opened 2022-08-29
metamask-ledger integration issue


hwwhww:
Reported by @fredriksvantes:

### Issue
My tests shows that metamask-ledger works fine on websites like app.uniswaporg, but not on the launchpad. i tested with two different OS and two different browsers on each OS, and the latest ledger firmware. it provides this error message:
"Ledger: Condition of use not satisfied (denied by user?) (0x6985) after pressing confirm in metamask.

### Possible root cause

Locally, i updated the web3, web3-utils, all @ethersproject and all `@web3-react packages` to their latest version, but it's still not working. my guess is something stopped working with the `@web3-react/injected-connector`, which can't be updated as it has been deprecated in favor of the "metamask" package: https://github.com/Uniswap/web3-react/tree/main/packages/metamask

---

Hi @wackerow, do you have bandwidth to help investigate how to switch our metamask package dependency?

I suppose it has to be fixed based on the node.js & web3 package upgrade fix (#543)

valefar-on-discord:
I played around with this for a few hours and am confident that updating the web-react package would likely fix the issue.

However, as mentioned by others, I too ran into dependency hell and was concerned about the blast radius of the change.

I had a brief discussion with Paul and we agreed the best path forward, given the state of this project, is to provide better error messaging to the user which I did #692 

Ethstaker sees this issue pop up quite often and we answer in the same manner that the error message describes so we are confident that, though it isn't a perfect solution, it will provide clear direction to the user allowing them to resolve this issue on their own.

PR #692, opened 2024-05-04
Add error message for user when ledger fails when attempting to sign

valefar-on-discord:
## Description

When a user attempts to perform a deposit with a Ledger device, if **Blind Signing** and **Debug Data** is not enabled, the transaction will fail with seemingly no reason. This has been a problem for years and still people ask about it on the EthStaker discord. Though this issue may be resolved by updating the web3-react library from v6 to v8, a brief attempt at doing so resulted in dependency hell and would be quite a drastic change for this one small issue. I had a quick chat with @wackerow about this and we agreed an error message is the best approach. 

The error message will appear if either **Blind Signing** or **Debug Data** are disabled or, sadly, if the user simply rejects from the Ledger device. We have no way to differentiate but I feel the benefit of the guidance outweighs the false positive.

I am a little uncertain on exact wording or if there is a stylistic standard around presenting error messages but happy to update as necessary. 

## Related Issues

- #549

### DIFF
--- a/src/pages/Transactions/index.tsx
+++ b/src/pages/Transactions/index.tsx
@@ -9,11 +9,13 @@ import { AbstractConnector } from '@web3-react/abstract-connector';
 import _every from 'lodash/every';
 import _some from 'lodash/some';
 import { DepositKeyInterface, StoreState } from '../../store/reducers';
+import { Alert } from '../../components/Alert';
+import { Button } from '../../components/Button';
 import { Heading } from '../../components/Heading';
+import { Link } from '../../components/Link';
 import { Paper } from '../../components/Paper';
 import { Text } from '../../components/Text';
-import { Button } from '../../components/Button';
-import { Link } from '../../components/Link';
+import { WorkflowPageTemplate } from '../../components/WorkflowPage/WorkflowPageTemplate';
 import { routesEnum } from '../../Routes';
 import { KeyList } from './Keylist';
 import { handleMultipleTransactions } from './transactionUtils';
@@ -21,7 +23,6 @@ import { TARGET_NETWORK_CHAIN_ID } from '../ConnectWallet/web3Utils';
 import { web3ReactInterface } from '../ConnectWallet';
 import { WalletDisconnected } from '../ConnectWallet/WalletDisconnected';
 import { WrongNetwork } from '../ConnectWallet/WrongNetwork';
-import { WorkflowPageTemplate } from '../../components/WorkflowPage/WorkflowPageTemplate';
 import {
   DepositStatus,
   DispatchTransactionStatusUpdateType,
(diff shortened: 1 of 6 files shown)


valefar-on-discord:
Ledger_error

src/pages/Transactions/transactionUtils.tsx
```diff
@@ -50,6 +50,19 @@ const isUserRejectionError = (error: any) => {
   return false;
 };
 
+const isLedgerError = (error: any) => {
+  if (error.code === -32603) {
+    // Error message when rejecting from ledger or blind signing and debug not enabled
+    if (
```
valefar-on-discord, line 56: Obviously could do `if (... && ...)` but followed commenting pattern of `isUserRejectionError`

valefar-on-discord:
Friendly ping on this. 

I am still assisting users with this issue and though this isn't the optimal solution, it will still help.

src/pages/Transactions/transactionUtils.tsx
wackerow: Hey @valefar-on-discord do you happen to know the error code that results from not having the proper Ledger settings enabled? Adding error handling here seems like a minimally invasive and safe improvement.

src/pages/Transactions/transactionUtils.tsx
```diff
@@ -50,6 +50,19 @@ const isUserRejectionError = (error: any) => {
...
+      error.message.includes('Ledger: Unknown error while signing transaction')
+    ) {
+      return true;
+    }
+  }
+
+  return false;
+};
```
wackerow, line 64: @valefar-on-discord Oops sorry, was on mobile and missed the context...

So `-32603` is the error code... is there a reason we need to also check for the "Ledger:" string below? If so, I may consider just checking for `.toLowerCase().includes("ledger")` so any error message with "Ledger" in it will be flagged.

And also, we could simplify this to just return `error.code === -32603 && error.message.toLowerCase().includes("ledger")` as a boolean:

```suggestion
// Error message when rejecting from ledger or blind signing and debug not enabled
const isLedgerError = (error: any) =>
  error.code === -32603 && error.message.toLowerCase().includes("ledger")  {
```

valefar-on-discord: Yup yup, sounds good.

Thanks for taking a look!

src/pages/Transactions/index.tsx
```diff
@@ -156,6 +157,21 @@ const _TransactionsPage = ({
...
+          k => k.transactionStatus === TransactionStatus.LEDGER_ERROR
+        ) && (
+          <Alert variant="error" className="mb20">
+            <FormattedMessage
+              defaultMessage="There was a problem trying to sign with your Ledger device. Please verify your Ledger
+              device has both {blindSigning} and {debugData} enabled. These can be accessed by selecting the Ethereum
+              application and then entering the Settings menu."
+              values={{
+                blindSigning: <b>Blind Signing</b>,
+                debugData: <b>Debug Data</b>,
```
wackerow, line 170: @valefar-on-discord Double checking here, do we want these to be translated? I'm not deeply familiar with any non-English Ledger interface. If we want these terms to be translated, we'll need to  wrap them in their own `FormattedMessage` component.

valefar-on-discord: That's a great question but I don't believe they are translated.

If you take a look at the [ledger documentation in a different language](https://support.ledger.com/hc/de/articles/4405481324433-Aktivieren-der-Blindsignatur-in-der-Ethereum-ETH-App?docs=true), it will show "Blind Signing"

valefar-on-discord: I'm going to update the instructions based on this little discovery.

valefar-on-discord: Updated to extract hard-coded words and used `<strong>`

valefar-on-discord:
Screenshot 2024-06-14 at 7 24 55 PM

Updated version

wackerow (approved):
Lgtm! Thanks @valefar-on-discord

PR #701, opened 2024-07-01
Update default testnet to Holesky, deprecate old testnets

wackerow:
## Description
- Removes any usage of Ropsten, Zhejiang or Goerli testnets (all deprecated)
- Updates Goerli references to use Holesky as the current default testnet


hwwhww (approved):
LGTM!

PR #714, opened 2024-10-26
Removing cli version checker

valefar-on-discord:
When generating a deposit with the [staking-deposit-cli](https://github.com/ethereum/staking-deposit-cli), the resulting deposit file will contain a cli version: `deposit_cli_version`. The launchpad currently verifies this value is above a certain threshold defined in the env file, 2.7.0. 

There are two problems:
- If this check fails, the resulting error presented to the user says there is a problem with the network
- This makes it difficult for multiple tools to exist

To get around these two issues, this PR removes the version check entirely.

As a side effect, `compare-version` is no longer a direct dependency as this was the only location where it it is used but it is still a dependency of `husky`

### DIFF
--- a/.env.holesky
+++ b/.env.holesky
@@ -4,7 +4,6 @@ export REACT_APP_ETH_REQUIREMENT=524288
 export REACT_APP_ETH_DEPOSIT_OFFSET=0
 export REACT_APP_TESTNET_LAUNCHPAD_NAME=Holesky
 export REACT_APP_GENESIS_FORK_VERSION=0x01017000
-export REACT_APP_MIN_DEPOSIT_CLI_VERSION=2.7.0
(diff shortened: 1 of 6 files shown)


valefar-on-discord:
@wackerow Here is the PR as promised.

wackerow (approved):
Given the switch to formally recommending the [eth-educators/ethstaker-deposit-cli](https://github.com/eth-educators/ethstaker-deposit-cli/) which has it's own versioning, it makes sense to remove this.

@CarlBeek @hwwhww Any concerns you have here? Otherwise going to bring this in

wackerow:
After quick chat with @CarlBeek, some agreement that this doesn't seem ideal, but will at least clear up the issues it's presenting. Will merge and we can continue to discuss potential stronger solutions for #719 as needed

PR #720, opened 2025-01-02
fix: update lodestar checklist and links

philknows:
This PR supersedes #707 and includes additional corrections and fixes to Lodestar's information on the Launchpad. This PR also adds Lodestar to display as part of the checklists for prospective stakers.


wackerow (approved):
Sorry for delay here @philknows... just committed a small patch, but looks good now!

image

PR #721, opened 2025-01-23
Pectra

wackerow:


CarlBeek (approved):
This looks great, @wackerow! I think we'll want to change some nits before mainnet, but this is perfect for the testnets

wackerow:
Agree! Pulling this in and will tee up a deploy to `master` for Holesky launch; can apply patches before any `mainnet` branch deploy

PR #722, opened 2025-01-23
fix: sha3 package build error

wackerow:
`sha3` indirect dependency tracing to `@web3-react/portis-connector` package no longer compatible, causing builds to break during dependency installation. removing package while still on older version of node to restore builds

Also patches a linting error from recent Nethermind page change, removing unused imports.


wackerow:
I've built this into my changes in #721 but we can merge it in isolation to `dev` to get that branch working again

README.md
```diff
@@ -26,7 +26,7 @@ The Infura Project ID will enable the network status progress bar showing the ba
 REACT_APP_INFURA_PROJECT_ID=your-infura-project-id-here
 ```
 
-Note: The Portis wallet option is not available when running locally
+Note: The Portis wallet option is currently not available
```
CarlBeek, line 29: I don't think we even really need to have this here, I think we can just nuke Portis & note something about removing it in a changelog & commit history

CarlBeek (approved):
Always love a commit that removes 10x the lines it adds, LGTM!

PR #724, opened 2025-01-25
Pectra Validator EL Actions

valefar-on-discord:
- Moved TransactionStatusModal from TopUp to be a generalized component
- TopOff update was because if a single validator is returned the api responds with an object instead of an array. This is either recent to the api or was not handled properly.


package.json
```diff
@@ -55,6 +55,7 @@
     "tiny-invariant": "^1.1.0",
     "typescript": "^3.8.3",
     "web3": "^1.7.5",
+    "web3-core": "^1.7.5",
```
valefar-on-discord, line 58: This is already a dependency it is now just explicit

wackerow:
Pulling these changes into `pectra` branch to gain access to Netlify preview deploys for these features

PR #726, opened 2025-01-31
Align with beaconcha.in validator status and remove execution credential filter during consolidation

valefar-on-discord:
We need to align with the validator status' from beaconcha.in when displaying certain information such as filtering for only active (online or offline) validators during consolidation requests. 

This contains a minor refactor to use these status' as an enum for consistency and during this refactor it was noted a few status' were not accounted for in ValidatorTable.

Additionally, the previous PR added a filter during consolidation for only execution addresses which was a mistake. You can consolidation with a source of either 0x01 or 0x02.

### DIFF
--- a/src/intl/compiled/en.json
+++ b/src/intl/compiled/en.json
@@ -275,6 +275,20 @@
       "value": "If at all possible, consider running another client at this time to help protect yourself and the network."
     }
   ],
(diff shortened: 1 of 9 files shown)


valefar-on-discord:
These changes were created from an older version. Will create a new one.

PR #727, opened 2025-01-31
fix: added user partial withdrawal transaction code

remyroy:
The pre-commit hook rearranged the intl json file content for some reason.

This should fix the partial withdrawal implementation.


wackerow:
All good with the intl json... it was probably just something that didn't get updated in the other branch.

wackerow (approved):

PR #731, opened 2025-02-06
Hide validator actions modals during transaction submission

valefar-on-discord:
Having a modal on top of another modal is not ideal and was noticing some UX issues such as a shaking screen and a disabled scrollbar during some instances.

Three options
1) Hide input modal during transaction submission and redisplay if transaction failed
2) Hide input modal during transaction submission and do not redisplay
3) Migrate transaction details into input modal as a second screen

This PR explores option 1


valefar-on-discord:
Going to go with option 3

Issue #718, opened 2024-12-18
Update client diversity warnings


CarlBeek:
According to [clientdiversity.org](https://clientdiversity.org/#distribution), the warnings provided for some of the clients are now outdated, specifically:

- geth is now below 66%
- Lighthouse is now below 33%
- Nethermind is now above 33%

We have two options on how to proceed, either we just update the warnings to reflect the latest client distributions, or we remove them all together. Personally given the lack of precision in the precise distribution, I am ok removing the warnings for Lighthouse & Prysm, as they are within error margins of being < 33%. The geth warning I could go either way on as they are still the dominant client by far.

wackerow:
Fixed with #732

PR #732, opened 2025-02-12
Remove diversity warnings

CarlBeek:
As discussed in #718, Geth no longer has >66% of node share and Prysm and Lighthouse are within the margin of error of <33%, so it is no longer necessary to have these warnings.


wackerow (approved):
Thanks @CarlBeek!

wackerow:
And thanks Ethereum community! ❤️

PR #739, opened 2025-02-24
feat: guard against EXCESS_INHIBITOR queue

wackerow:
Fixes bug causing browsers to freeze pre-Pectra-fork since these contracts are returning the 2^255 EXCESS_INHIBITOR value when fetching the queue lengths. This falls back if this is returned and uses a queue length of 0. Will only impact prior to fork.

### DIFF
--- a/src/pages/Actions/utils.ts
+++ b/src/pages/Actions/utils.ts
@@ -6,6 +6,7 @@ import {
   COMPOUNDING_CONTRACT_ADDRESS,
   WITHDRAWAL_CONTRACT_ADDRESS,
   TICKER_NAME,
+  EXCESS_INHIBITOR,
 } from '../../utils/envVars';
 
 export type Queue = {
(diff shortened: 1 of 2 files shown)


src/pages/Actions/utils.ts
```diff
@@ -35,13 +36,23 @@ const getRequiredFee = (queueLength: BigNumber): BigNumber => {
...
-    0 // EXCESS_WITHDRAWAL_REQUESTS_STORAGE_SLOT
-  );
-
-  if (!queueLengthHex)
-    throw new Error('Unable to get compounding queue length');
+  let queueLengthHex;
+  try {
+    queueLengthHex = await web3.eth.getStorageAt(
+      COMPOUNDING_CONTRACT_ADDRESS!,
+      0 // EXCESS_WITHDRAWAL_REQUESTS_STORAGE_SLOT
```
remyroy, line 43: Minor stuff, but this should be `EXCESS_CONSOLIDATION_REQUESTS_STORAGE_SLOT` instead here.

wackerow: Good spot, will note this and patch separate before mainnet

PR #746, opened 2025-03-18
feat: import approved/compiled intl strings

wackerow:
## Description
- Imports all approved intl strings from Crowdin (compiled form)
- Includes all initial Pectra copy updates


CarlBeek (approved):
From the few spot checks I did, this seems good to me

PR #751, opened 2025-04-25
Pectra mismatch consolidation

valefar-on-discord:
Adds ability to consolidate a validator into one that is not associated with the connected wallet. This is done by searching the validator dropdown by a pubkey. If one is discovered, it will be selected.  Also handling messaging if the discovered validator does not have 0x02 credentials. A little safety check is also being added in this case where the user will need to type the validator index of what they are sending to. This can be easily updated if a more strenuous message is desired.

This is branched off of #750 so that should be landed first

### DIFF
--- a/src/pages/Actions/components/PushConsolidation.tsx
+++ b/src/pages/Actions/components/PushConsolidation.tsx
@@ -155,7 +155,7 @@ const PushConsolidation = ({
     }
   }, [targetValidator]);
 
-  const validValidatorState = useMemo(() => {
+  const validValidatorStatus = useMemo(() => {
     // Handle both dora and beaconcha.in status'
     return (
       targetValidator &&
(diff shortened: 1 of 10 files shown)


remyroy (changes requested):
It seems like this is fine from a feature point of view. There are a few things I noticed that could be improved I think.

On a 1920 x 1080 resolution, the modal screen showing the consolidation request does not fit all the content and the layout is somewhat screwed. Using a vertical scrollbar on the modal part or making the content slightly smaller in height could improve this situation.
Screenshot from 2025-04-29 11-13-57

It would be nice to be able to enter a validator index as the target validator to transfer to in addition to the public key. Right now, you can only enter the public key to choose your target validator.

The confirmation method of retyping only the target validator index is not strong enough for the potential consequences to me. I would prefer having to retype a longer sentence such as: "I TRANSFER MY FULL VALIDATOR BALANCE TO THE OWNER OF {INDEX}" for a stronger confirmation.

Some explorers are still having issue with display this transaction. Here is mine:

- Transaction: https://hoodi.etherscan.io/tx/0x6e11904576df4510a1c65ba966082e37ba66c9049bc06f97c0ba7f992b0ec13f
- Source validator (Dora): https://dora.hoodi.ethpandaops.io/validator/1086489?v=consolidationrequests
- Source validator (Beaconcha.in): https://hoodi.beaconcha.in/validator/1086489#consolidations
- Target validator (Dora): https://dora.hoodi.ethpandaops.io/validator/1074380?v=consolidationrequests
- Target validator (Beaconcha.in): https://hoodi.beaconcha.in/validator/1074380#consolidations

valefar-on-discord:
Thanks for taking a look @remyroy 

Here is what I've done:
- I've updated the confirmation message as suggested
- I've fixed the scrolling issues that existed in both pull/push modals
- If there was only 1 0x02 validator associated with the account and it was selected, the push flow was disabled. I removed this check
- If there was only 1 0x02 validator and the user selected a 0x01 validator to push, the 0x02 validator would get auto selected every 30 seconds even if the user searched for a validator by pubkey. I added a param to the ValidatorSelect component to prevent this from happening which is only used in the Push flow.

The only thing I haven't done was the validator index search as it adds a bit of complexity: When searching by pubkey there is a very strict length that can be used to know when to search. When the search completes, if any is discovered, the discovered validator will be auto-selected, the component blurs, and the user can complete the flow. With an index search, even with a debounce, it can be problematic unless there is a copy-paste. 

Additionally, you could make the argument that if the user knows the pubkey and confirms the validator index the likelihood they are making a mistake is much less. A typo of a validator index on the other hand seems much more plausible. 

Lemme know what you think

src/pages/Actions/components/PushConsolidation.tsx
```diff
@@ -123,12 +135,46 @@ const PushConsolidation = ({
...
+    if (targetValidator) {
+      return (
+        +targetValidator.withdrawalcredentials.slice(0, 4) >=
+        +COMPOUNDING_CREDENTIALS
+      );
+    }
+  }, [targetValidator]);
+
+  const validValidatorState = useMemo(() => {
+    return targetValidator && targetValidator.status === 'active_ongoing';
```
remyroy, line 159: Dora and Beaconcha.in APIs might differ here but I think this one should be testing for (`active_online` or `active_offline`) for beaconcha.in

```
return targetValidator && (targetValidator.status === 'active_online' || targetValidator.status === 'active_offline');
```

remyroy (changes requested):
Everything seems good with the new changes except for this one condition I just noticed.

src/pages/Actions/components/PushConsolidation.tsx
valefar-on-discord: Great catch! Given the addition of dora support here is minimal we can add it for now though there are obviously other areas around the app that don't work with dora 🫠

valefar-on-discord: Resolved

src/pages/Actions/components/PushConsolidation.tsx
```diff
@@ -123,12 +135,50 @@ const PushConsolidation = ({
...
+        +COMPOUNDING_CREDENTIALS
+      );
+    }
+  }, [targetValidator]);
+
+  const validValidatorState = useMemo(() => {
+    // Handle both dora and beaconcha.in status'
+    return (
+      targetValidator &&
+      ['active_online', 'active_ongoing'].includes(targetValidator.status)
```
remyroy, line 162: I think we need to include `active_offline` in there too. From my understanding, `active_online` is when a validator is active and actively validating and `active_offline` is when a validator is active but failed to produce attestation recently. In both cases, they should be valid targets for this push. So the valid states should be `['active_online', 'active_offline', 'active_ongoing']`

remyroy: Everything else is good on my side.

CarlBeek:
Just looking at the modal @remyroy was talking about [above](https://github.com/ethereum/staking-launchpad/pull/751#pullrequestreview-2804134838), we should have a check on the balance that won't allow you to consolidate into a validator if the result will be a balance > MAX_EB (2048 ETH).

src/pages/Actions/components/PushConsolidation.tsx
CarlBeek: I agree about including `active_offline` here, good catch @remyroy.

I've checked the Beaconcha.in codebase and I am comvinced this is sufficient, but I was unable to confirm these are the correct statuses in Dora.

valefar-on-discord: I took a look at dora. It uses [go-eth2-client states](https://github.com/attestantio/go-eth2-client/blob/07599e7891634b175c498ea5c5f842a19bc29a61/api/v1/validatorstate.go#L53) which doesn't seem to differentiate between active/offline like beaconcha.in.

src/pages/Actions/components/PushConsolidation.tsx
```diff
@@ -245,253 +367,298 @@ const PushConsolidation = ({
...
+                    gap: '0.5rem',
+                  }}
+                >
+                  <label
+                    htmlFor="confirm-push-input"
+                    style={{ marginBottom: '0.25rem' }}
+                  >
+                    <Text>
+                      <FormattedMessage
+                        defaultMessage="Please type {boldConfirmationMessage} to acknowledge you understand where your validator balance is being transferred to."
```
wackerow, line 608: > ...to acknowledge you understand where your validator balance is being transferred to

Consider "...to acknowledge you understand your validator is changing ownership"

src/pages/ConnectWallet/web3Utils.ts
wackerow, line 1: Would revert these changes before merging

wackerow (changes requested):
Overall looks good. 

Tried it out... any account that isn't 0x02 is ignored entirely.

I'd personally advocate we leave it as a pubKey search only, and not allow for indices. Given the rapid nature of these updates, I would err on the side of more friction here.

- [ ] Requesting we remove the devnet6 listing before merging.

Otherwise I pulled in latest from the target pectra branch to clear conflicts, and updated the confirmation string here to be intl friendly (since we need to get all this translated for other languages).

Left one non-blocking copy suggestion

src/pages/ConnectWallet/web3Utils.ts
valefar-on-discord: ack

src/pages/Actions/components/PushConsolidation.tsx
valefar-on-discord: Updated

valefar-on-discord:
Thanks for taking a look @wackerow 

- Removed devnet6 config changes
- Updated confirmation message
- Resolved merge conflicts
- Also I updated the 0x01 pushing warning message to bold `Upgrade Account`

Ran through some tests without seeing any major issues. There are a few UX items that could be improved such as what has been suggested and if the user is trying to push a 0x01 validator into itself, it should work but currently doesn't. I think we are all in agreement, especially due to time constraints, we should move forward with what we have.

wackerow (approved):
Pulling into [`pectra-post-holsesky`](https://github.com/ethereum/staking-launchpad/pull/744) PR branch

PR #755, opened 2025-05-02
Add dora validator statuses

remyroy:
The Dora API uses different statuses than those from beaconcha.in. It uses the one from https://github.com/attestantio/go-eth2-client . The source can be found on https://github.com/attestantio/go-eth2-client/blob/07599e7891634b175c498ea5c5f842a19bc29a61/api/v1/validatorstate.go#L49

This PR add the additional statuses so we can use both sets of APIs. This is mostly for visual support for the staking-launchpad.


wackerow (approved):
lgtm!

PR #749, opened 2025-03-28
fix: use a working el explorer for hoodi

remyroy:
- Replaced the existing configured hoodi explorer with a working one. The EthPandaOps explorer is deprecated and the etherscan one is likely going to be more stable and reliable on Hoodi.

PR #745, opened 2025-03-18
Add Hoodi config to network list

wackerow:
## Description
- Adds the Hoodi network configuration for wallet connection
- Reverts Hoodi env var overrides, as the Hoodi env var configuration is currently being applied to `dev` and (to my knowledge) this feature branch
- Adds `.env.hoodi`, updates `.env.template` and `src/utils/envVars.ts` to use Hoodi over Holesky

## Preview link
https://deploy-preview-745--dapper-rolypoly-9814ad.netlify.app/

PR #744, opened 2025-03-03
Pectra touch-ups

wackerow:

PR #756, opened 2025-05-05
`dev` -> `master` ahead of Pectra fork

CarlBeek:
### Headliner:
 - #751 - Fully Featured Pectra Consolidations

### Supporting Cast:
 - #744 - Feature branch for the below changes
 - #745 - Adds Hoodi testnet preset
 - #746 - Adds intl/translations for Pectra features
 - #748 - Fixes config error in Hoodi preset
 - #749 - Switch to Etherscan for Hoodi EL explorer
 - #750 - Addresses potential issues with txs not having sufficient fees
 - #752 - Correctly applies strong font weights
 - #753 - Handles bug due to different return types from RPC
 - #755 - Correctly handles validator statuses from Dora API
 
 A big shout out to @valefar-on-discord, @remyroy, & @wackerow for pushing to get these features ready!


wackerow (approved):

.env.template
```diff
@@ -39,28 +39,22 @@ REACT_APP_TEKU_INSTALLATION_URL=https://docs.teku.consensys.io/get-started/insta
...
 # number
 REACT_APP_ETH_DEPOSIT_OFFSET=0
 # string
-REACT_APP_FAUCET_URL=https://www.holeskyfaucet.io/
-# string
-REACT_APP_RPC_URL=https://rpc.holesky.ethpandaops.io
+REACT_APP_FAUCET_URL=https://hoodi-faucet.pk910.de/
 # string
-REACT_APP_BEACONCHAIN_URL=https://holesky.beaconcha.in
+REACT_APP_BEACONCHAIN_URL=https://dora.hoodi.ethpandaops.io/
```
remyroy, line 56: Is this used anywhere in terms of deployment? If so, we might wanna switch to using `https://hoodi.beaconcha.in` instead here. Even if we want to use Dora, we would need to remove the leading `/` to make it work.

remyroy (changes requested):
I found a minor configuration issue.

.env.template
wackerow: No, this is just the template for users to in their local environments... can patch now separately in a small PR but don't think this needs to block.

wackerow (approved):

.env.template
CarlBeek: Addressed in #757

Issue #760, opened 2025-05-10
Topping up should not be possible with an invalid amount of ether


julian-st:
Topping up after pectra is possible with values like 1.217792 but the transactions fails with the error message: DepositContract: deposit value not multiple of gwei

So the launchpad should prevent the user from putting values that are not multiple of gwei

PR #764, opened 2025-05-13
[#760] Added rounding down of available ether to valid number for dep…

julian-st:
Added rounding down of available ether to valid number for deposit contract

see #760


src/pages/Actions/components/AddFunds.tsx
```diff
@@ -209,10 +209,11 @@ const AddFunds = ({ validator }: AddFundsProps) => {
...
   const primaryLabel = <FormattedMessage defaultMessage="Stake more funds" />;
 
-  const maxValue = BigNumber.min(
-    etherMaxAvailableToSend,
-    etherMaxUseful
-  ).toNumber();
+  const maxValue = BigNumber.min(etherMaxAvailableToSend, etherMaxUseful)
+    .multipliedBy(1e8)
+    .integerValue(BigNumber.ROUND_DOWN)
+    .div(1e8)
```
wackerow, line 215: Hey @julian-st! Thanks for the PR... Few things

For gwei precision we'd want 9 here, and should be able to just use `.decimalPlaces(9, BigNumber.ROUND_DOWN)`

But, this issue seems to be primarily addressed in the `createAddFundsTransaction` function:

https://github.com/julian-st/staking-launchpad/blob/0bade1b9f8796fb668e412abc091689fcebb4a6f/src/pages/Actions/components/AddFunds.tsx#L164-L208

```tsx
  const createAddFundsTransaction = async () => {
    if (!etherAmount || !account) return;

    setTxStatus('waiting_user_confirmation');
    setShowTx(true);

    const walletProvider: provider = await (connector as AbstractConnector).getProvider();
    const web3: Web3Instance = new Web3(walletProvider);
    const contract = new web3.eth.Contract(
      contractAbi,
      DEPOSIT_CONTRACT_ADDRESS
    );
    const bnInput = new BigNumber(etherAmount);
    const transactionAmount = bnInput
      .multipliedBy(1e18) // <-- Change to 1e9?
      .integerValue(BigNumber.ROUND_DOWN); // Convert to Wei and ensure it's an integer
    const reconstructedRootAmount = bnInput
      .multipliedBy(1e9)
      .integerValue(BigNumber.ROUND_DOWN); // Convert to Gwei and ensure it's an integer

    const transactionParameters: SendOptions = {
      gasPrice: web3.utils.toHex(await web3.eth.getGasPrice()),
      from: account as string,
      value: transactionAmount.toFixed(),
    };

    const reconstructedKeyFile = {
      pubkey: bufferHex(validator.pubkey.substring(2)),
      withdrawalCredentials: Buffer.alloc(32),
      amount: reconstructedRootAmount,
      signature: Buffer.alloc(96),
    };

    const byteRoot = depositDataContainer.hashTreeRoot(reconstructedKeyFile);
    const reconstructedDepositDataRoot = `0x${buf2hex(byteRoot)}`;

    contract.methods
      .deposit(
        reconstructedKeyFile.pubkey,
        reconstructedKeyFile.withdrawalCredentials,
        reconstructedKeyFile.signature,
        reconstructedDepositDataRoot
      )
      .send(transactionParameters)
```

Marked one line with `// <-- Change to 1e9?` above, which maybe @valefar-on-discord or @remyroy could weigh in on, but I believe that would fix any issues where values with deeper precision than "gwei" are let through to the transaction.

julian-st: @wackerow thanks for the input, I changed it and used the function decimalPlaces

wackerow: Okay sure, but what about the rest of my comment? I think that would be the cleaner approach

julian-st: I changed also the decimal value for the transactionAmount. If I changed it to 1e9 wouldn't that convert it to Gwei and change the value down the line and in further handling of the value?

julian-st:
@valefar-on-discord thanks I used your changes for the handleValueChange function to have a better precision check

valefar-on-discord (approved):
Happy with this but I'd like @wackerow to sign off as well.

src/pages/Actions/components/AddFunds.tsx
```diff
@@ -142,10 +142,14 @@ const AddFunds = ({ validator }: AddFundsProps) => {
...
-      MIN_DEPOSIT_ETHER
-    );
+    let newValue = Math.max(Math.min(value, etherMaxUseful), MIN_DEPOSIT_ETHER);
+
+    // Do not allow more than 1 gwei precision as required by the deposit contract
+    newValue = new BigNumber(value)
+      .multipliedBy(ETHER_TO_GWEI)
+      .integerValue(BigNumber.ROUND_DOWN)
+      .dividedBy(ETHER_TO_GWEI)
+      .toNumber();
```
wackerow, line 152: We're immediately mutating the `newValue` variable, and unconditionally, so the initial `let` assignment is not needed, and we can use `const` instead. The multiply => integer rounding => dividing can be equivilently done with `.decimalPlaces(9, BigNumber.ROUND_DOWN)`, so I would suggest this:

```suggestion
    // Round to gwei precision as required by the deposit contract
    const newValue = new BigNumber(value)
      .decimalPlaces(9, BigNumber.ROUND_DOWN)
      .toNumber();
```

wackerow (dismissed):
This functionally looks good! Just a clean up request

src/pages/Actions/components/AddFunds.tsx
julian-st: @wackerow thanks for the input, I changed it

julian-st:
I fixed the cleanup

noelportillo (approved):

noelportillo (approved):

valefar-on-discord (approved):

PR #766, opened 2025-05-21
fix: top-up page max-EB correctly distinguishes compounding accounts

emalaj:
The Top Up tab incorrectly shows a Max EB of 32 even for compounding accounts because `validator.withdrawalcredentials` from `beaconcha.in` returns a long address (prepended with `0x02`) which `ValidatorTable.tsx:207` was not properly handling. Rather than fixing that hardcode, use the existing `utils.validators.getMaxEB`

The other `src/intl` changes were autogen-ed


wackerow (approved):
Looks good! Thanks @emalaj!

PR #771, opened 2025-06-06
feat: Update withdrawals page for Pectra upgrade - Add Type 2 (0x02) …

jmcpheron:
Update staking withdrawals documentation for Pectra upgrade

This PR updates the staking withdrawals page to accurately reflect how withdrawals work after the Pectra upgrade (May 2025).

Key changes:
- Clarified the difference between Type 1 and Type 2 validators
- Fixed incorrect information about automatic reward sweeps
- Added new withdrawal options available with Pectra
- Updated tool versions and links to support Type 2 validators

The previous documentation stated that manual withdrawal of specific amounts was not possible. This has been updated to reflect that Pectra now enables manual partial withdrawals for Type 2 validators through the new partialWithdraw() function, giving stakers more control over their funds.

### DIFF
--- a/src/pages/TopUp/components/ValidatorTable.tsx
+++ b/src/pages/TopUp/components/ValidatorTable.tsx
@@ -21,11 +21,10 @@ import { Paper } from '../../../components/Paper';
 import { Link } from '../../../components/Link';
 import {
   BEACONCHAIN_URL,
-  MIN_ACTIVATION_BALANCE,
   TICKER_NAME,
   ETHER_TO_GWEI,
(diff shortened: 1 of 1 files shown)


noelportillo (approved):

noelportillo (approved):

wackerow (approved):
Looks great, thanks @jmcpheron!

Pushed a commit with a few small adjustments. We're pointing everything to the ethstaker-deposit-cli tool now, so I replaced with that label. Refactored to just use the variables here wherever I could which doesn't change anything on how this will display. Otherwise looks good! Thanks for the help, pulling this in!

PR #797, opened 2025-12-17
Add missing title and meta tags to index.html

minimalsm:
## Summary
- Add default `<title>` tag for SEO crawlers
- Add `<meta name="description">` tag
- Add `og:title` and `og:description` meta tags
- Add `twitter:title` and `twitter:description` meta tags

This fixes the "Title tag missing or empty" issue flagged by AHREFS on:
- launchpad.ethereum.org/
- launchpad.ethereum.org/en/
- hoodi.launchpad.ethereum.org/
- holesky.launchpad.ethereum.org/

The issue is that crawlers don't execute JavaScript, so they see the raw HTML which had no title. React Helmet sets titles client-side, but this default ensures crawlers see something.

## Test plan
- [ ] Verify title appears in page source before JS loads
- [ ] Verify React Helmet still overrides with localized content

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -21,10 +21,12 @@
       };
     </script>
     <meta charset="utf-8" />
+    <title>Staking Launchpad</title>
     <meta
       name="viewport"
       content="width=device-width, initial-scale=1.0, maximum-scale=1, minimum-scale=1"
     />
+    <meta name="description" content="Become a validator and help secure the future of Ethereum." />
     <link
       rel="apple-touch-icon"
       sizes="180x180"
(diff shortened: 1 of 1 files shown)


valefar-on-discord (commented):
A concern that this change creates is we will now have multiple tags with differing information. 

Most social media crawlers do not execute javascript where the dynamic tags will not be discovered but this is not the case with search engine bots like Google and Bing. They will render before indexing.

I see three routes:
- Merge static & dynamic tags to avoid having multiple instances: This is very likely more work than it's worth
- Define which crawler bots we care most about and optimize for their limitations: Either ignore this change or delete the existing dynamic implementation
- Ignore multiple tag concern and we are good to go

PR #798, opened 2026-01-16
pkg JSON update

JAnalyticsAI:
Updating all dependencies in package.json in root


valefar-on-discord:
Appreciate the optimism but some of these packages are almost 6 years out of date and upgrading would result in a tremendous amount of breaking changes.

PR #799, opened 2026-03-04
refactor: use bearer auth for beaconcha.in fetch

wackerow:
## Summary
- Switch beaconcha.in API authentication from `apikey` query parameter to `Authorization: Bearer` header
- Clean up URL construction using `new URL()` instead of string interpolation

## Context
API calls through the Netlify proxy function were failing in production despite correct env var configuration. The beaconcha.in docs support both query param and Bearer token auth -- switching to the header approach to resolve the issue.

## Test plan
- [ ] Verify landing page loads network stats (total staked, validator count, APR)
- [ ] Verify TopUp page fetches validators by connected wallet
- [ ] Verify Actions page fetches validators by withdrawal credentials
- [ ] Check Netlify function logs for any auth errors

### DIFF
--- a/netlify/functions/beaconchain-api.js
+++ b/netlify/functions/beaconchain-api.js
@@ -18,18 +18,17 @@ exports.handler = async function(event) {
     };
   }
 
-  const BEACONCHAIN_API_KEY = process.env.BEACONCHAIN_API_KEY || '';
+  const apiKey = process.env.BEACONCHAIN_API_KEY || '';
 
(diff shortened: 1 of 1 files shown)


CarlBeek:
LGTM! API calls are back up

For some reason, not all the fields load on the home-page, but that's a separate issue. Everything else seems to work again.

Screenshot 2026-03-05 at 10 43 04 AM

CarlBeek (approved):

PR #802, opened 2026-03-04
temp: disable broken netlify server function API dependency

wackerow:
## Description
- temporarily disable API connections requiring netlify server functions until node version issues resolved
- displays banner to users explaining the these functions are currently not available and funds are safe

## TODO
- [x] Translate new strings


myelinated-wackerow (approved):
**Verdict: Approve**

Clean, focused emergency patch. The kill-switch approach is appropriate for a temporary fix, the banner messaging is clear and reassuring, and translations are complete across all 20 locales.

**One minor gap to be aware of:**

`src/pages/UploadValidator/validateDepositKey.ts` -- the `getExistingDepositsForPubkeys()` function (line 98) still calls `getBeaconchainV1ApiUrl()` and will hit the broken Netlify function endpoint during the deposit file upload flow. However, this is already handled gracefully: on failure, it dispatches `BeaconChainStatus.DOWN` to Redux (UploadValidator/index.tsx:220-222), and the user can still proceed with their deposit. The only degradation is that the **double-deposit safety check** (which warns users if their pubkeys already have on-chain deposits) will be unavailable. This seems acceptable for a temporary patch, but worth noting -- users uploading deposit files during the outage won't get the duplicate deposit warning.

Everything else looks good -- the `useBeaconchainData` hook, WithdrawalCredentials, TopUp, Actions, and NetworkStatus pages are all properly guarded.

**Security review -- no concerns found:**

- **No XSS risk:** The `ApiOutageBanner` component uses `FormattedMessage` with static, pre-compiled i18n strings only. No user input is interpolated into the banner. No use of `dangerouslySetInnerHTML`.
- **No information leakage:** The banner text is generic ("service update") and does not expose internal details about the Netlify/Node version issue, infrastructure topology, or error stack traces.
- **No new external requests or dependencies introduced:** The kill switch prevents outbound API calls; it does not add any new ones. No new packages added.
- **No credential or secret exposure:** The `BEACONCHAIN_API_DISABLED` flag is a hardcoded boolean (`true`), not derived from environment variables or secrets. The existing `BEACONCHAIN_URL` env var (used in the NetworkStatus fallback link) was already public and present before this PR.
- **Kill switch is code-level, not config-level:** Requires a deploy to toggle, which is appropriate for this kind of temporary emergency patch. No risk of accidental re-enablement via misconfigured env vars.
- **Deposit flow integrity preserved:** The core deposit workflow (key generation, BLS validation, on-chain transaction) is unaffected. Only the beaconchain API-dependent read operations (validator lookup, network stats, double-deposit check) are disabled. Users can still stake safely.

---

Reviewed by Claude (claude-opus-4-6)

wackerow:
Closing in lieu of #799 and Netlify env var config updates

